PR #5, opened 2023-02-23
fix: adding vitest and initial tagging function [TOL-909]

aodhagan-cf:
Adding initial tagging functionality

### DIFF
--- a/src/features/tagging/getLivePreviewProps.spec.ts
+++ b/src/features/tagging/getLivePreviewProps.spec.ts
@@ -1,0 +1,21 @@
+import { indexForPath, routeForIndex, StaticRoutes } from './staticRoutes';
+import { describe, it, expect } from 'vitest';
+import { getLivePreviewProps } from './getLivePreviewProps';
+
+describe('getLivePreviewProps', () => {
+  it('returns the expected props with a given entryId, fieldId and locale', () => {
+    const entryId = 'test-entry-id';
+    const fieldId = 'test-field-id';
+    const locale = 'test-locale';
+
+    const result = getLivePreviewProps({
+      entryId,
+      fieldId,
+      locale,
+    });
+
+    expect(result).toStrictEqual({
+      'data-contentful-field-id': `${entryId}:${fieldId}:${locale}`,
+    });
+  });
+});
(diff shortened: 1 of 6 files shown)


src/features/tagging/getLivePreviewProps.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+    const fieldId = 'test-field-id';
+    const locale = 'test-locale';
+
+    const result = getLivePreviewProps({
+      entryId,
+      fieldId,
+      locale,
+    });
+    expect(result).toStrictEqual({
+      'data-contentful-field-id': `${entryId}:${fieldId}:${locale}`,
```
YvesRijckaert, line 16: Just thinking out loud here, what if the user has a ':' inside of the fieldId? Won't this cause issues to extract the variables from the string?
We had a similar issue with the path in Compose before

YvesRijckaert (approved):
Small question

src/features/tagging/getLivePreviewProps.spec.ts
aodhagan-cf: Yeah defo a good point! I didn't want to reintroduce the 👺 or what we had before. Can you think of a more sensible delimiter?

aodhagan-cf: I guess another way to do it could be

```
'data-contentful-field-id': fieldId,
'data-contentful-entry-id': entryId,
'data-contentful-locale': locale`,
```

then no need for the delimiting

YvesRijckaert: Maybe a delimiter that is less likely to appear within the variable values? Like the pipe symbol (|) or  tilde symbol (~)? Not sure...

YvesRijckaert: Ah yes I like your approach 👏

chrishelgert: Just one note:
That could be resolve to a lot of markup on the website and bloat the HTML response (on server-side-rendering).
I guess we should give some best practices to customers to only render the tags on a preview system so they don't affect the performance of their website?

aodhagan-cf: @chrishelgert yeah makes sense!

PR #6, opened 2023-02-23
fix: adding types to build [TOL-909]

aodhagan-cf:
Just realised when attempting to use the package that types are not exported 🙃


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/live-preview",
-  "version": "1.0.1",
```
aodhagan-cf, line 3: Github is not automatically doing this, must be something wrong with the setup 🤔

chrishelgert: mhh that strange 🤔

chrishelgert: we need this plugin for updating the package.json version:
https://github.com/semantic-release/git

aodhagan-cf: I think this is already in the package.json

chrishelgert (approved):

PR #14, opened 2023-02-28
feat: add tooltip bttn [TOL-937]

YvesRijckaert:
Add init function and tooltip button

I restructured the code a bit: we now have a ContentfulLivePreview class with an init function and the getProps function (to add the data-attributes). The init function creates a new FieldTagging class that is responsible for the field tagging feature.


src/tests/resolveIncomingMessage.spec.ts
```diff
@@ -0,0 +1,36 @@
+// @vitest-environment jsdom
+import { describe, test, expect, beforeEach, afterEach, vi } from 'vitest';
+import LivePreview from '../live-preview';
+
+describe('resolveIncomingMessage', () => {
```
YvesRijckaert, line 5: I will add more tests as part of a different PR (not to bloat this PR)

src/field-tagging.ts
```diff
@@ -0,0 +1,127 @@
...
+      this.tooltip.style.top = upperBoundOfTooltip + 'px';
+      this.tooltip.style.left = left + 'px';
+
+      return true;
+    }
+
+    return false;
+  }
+
+  private addStyleOnHover(e: MouseEvent) {
```
chrishelgert, line 66: it adds the popover on hover to this element right? not really styles 🤔

chrishelgert (commented):
Looks good 🌻 
I am not a big fan of classes but I think it makes sense in this case, I added some questions/maybe improvements.
General question would it make sense to have an example inside the repo that can be deployed on the PR?

chrishelgert (approved):

PR #25, opened 2023-03-17
fix: prevent warning when called with undefined or null during develo…

chrishelgert:
fix: prevent warning when called with undefined or null during development.

useDeepCompareEffect has an integrated check if the provided dependencies are primitives during development.
As the provided data can be `null` or `undefined` during loading, it will throw an error which can crash the Developer Experience.

### DIFF
--- a/src/react.ts
+++ b/src/react.ts
@@ -1,22 +1,22 @@
 import { useState } from 'react';
 
-import useDeepCompareEffect from 'use-deep-compare-effect';
+import { useDeepCompareEffectNoCheck } from 'use-deep-compare-effect';
 
 import { ContentfulLivePreview } from '.';
-import { Entity } from './types';
+import { Argument } from './types';
 import { debounce } from './utils';
 
-export function useContentfulLiveUpdates<T extends Entity | null | undefined>(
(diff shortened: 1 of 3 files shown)


src/react.ts
```diff
@@ -3,10 +3,10 @@ import { useState } from 'react';
 import useDeepCompareEffect from 'use-deep-compare-effect';
 
 import { ContentfulLivePreview } from '.';
-import { Entity } from './types';
+import { Argument } from './types';
 import { debounce } from './utils';
 
-export function useContentfulLiveUpdates<T extends Entity | null | undefined>(
+export function useContentfulLiveUpdates<T extends Argument | null | undefined>(
   data: T,
```
aodhagan-cf, line 10: how about `data: T = {}` to fallback to an empty object?

I could be wrong but
```
}, [data || {}]);
```

looks like it's creating a new object on every render

aodhagan-cf (approved):

src/react.ts
chrishelgert: Sadly not working, because default parameters only are used on `undefined` but not for null :( 
but we can use `useDeepCompareEffectNoCheck`

PR #26, opened 2023-03-17
chore: define multiple exports

chrishelgert:


package.json
```diff
@@ -21,7 +33,7 @@
   },
   "scripts": {
     "start": "vite build --watch",
-    "build": "tsc && vite build",
```
aodhagan-cf, line 24: Don't actually know if we need tsc here as the plugin generates the types for us 🤔

src/index.ts
```diff
@@ -4,6 +4,8 @@ import { LiveUpdates } from './live-updates';
 import { Entity, LivePreviewProps, SubscribeCallback, TagAttributes } from './types';
 import { sendMessageToEditor } from './utils';
 
+export * from './react';
```
aodhagan-cf, line 7: I've included react in the index bundle... would love to not do that

package.json
```diff
@@ -6,12 +6,28 @@
...
   "files": [
-    "dist"
+    "dist",
+    "react",
+    "style.css"
   ],
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
```
chrishelgert, line 20: we could add the "types": "./dist/index.d.ts", as it should now  work since ts@4.9

aodhagan-cf: Does that mean that all of our customers need to be on that version of ts?

chrishelgert: mh you're right that would be the requirement

aodhagan-cf:
Hey @chrishelgert, is anything else needed on this PR?

chrishelgert:
> Hey @chrishelgert, is anything else needed on this PR?

@aodhagan-cf no should be ready to ship 🚢

YvesRijckaert (approved):
Let's create a follow up to update the live-preview package in the marketing template + REST template?

chrishelgert:
> Let's create a follow up to update the live-preview package in the marketing template + REST template?

Can do this directly once it is published :)

2wce (approved):

PR #31, opened 2023-03-21
feat: live update primitive field values

z0al:
Todo:
- [ ] Fetch Content types from Host on the initial load
- [ ] More testing

### DIFF
--- a/src/graphql/entries.ts
+++ b/src/graphql/entries.ts
@@ -8,7 +8,7 @@ import { isPrimitiveField, logUnrecognizedFields } from './utils';
  * @param contentType entity
  * @param data the GraphQL response to be updated
  * @param update CMA entry object containing the update
- * @param locale locale code
+ * @param locale code
  */
 export function updateEntry(
   contentType: ContentTypeProps,


src/graphql/entries.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { isPrimitiveField, logUnrecognizedFields } from './utils';
+
+/**
+ * Updates GraphQL response data based on CMA entry object
+ *
+ * @param contentType entity
+ * @param data the GraphQL response to be updated
+ * @param update CMA entry object containing the update
+ * @param locale locale code
```
YvesRijckaert, line 11:
```suggestion
 * @param locale code
```

YvesRijckaert (approved):
💯

src/live-updates.ts
```diff
@@ -16,9 +17,11 @@ export class LiveUpdates {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   private mergeGraphQL(initial: Argument, locale: string, updated: Entity): Argument {
-    // TODO: https://contentful.atlassian.net/browse/TOL-1000
-    // TODO: https://contentful.atlassian.net/browse/TOL-1022
-    return initial;
+    if ((initial as any).__typename === 'Asset') {
```
chrishelgert, line 20: `Argument` can also be an array currently, maybe we should adjust this check here

src/graphql/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  for (const field of Object.keys(data)) {
+    if (!recognized.has(field)) {
+      console.warn(`Unrecognized field '${field}'. Note that GraphQL aliases are not supported`);
+    }
+  }
+}
+
+export function isPrimitiveField(field: ContentFields): boolean {
+  const types = new Set(['Symbol', 'Text', 'Integer', 'Boolean', 'Date', 'Location', 'Object']);
```
aodhagan-cf, line 17: nit: I would extract this out of the function so you don't have to instantiate it even function call

src/graphql/entries.ts
```diff
@@ -0,0 +1,39 @@
...
+    fields.map((f) => f.apiName ?? f.name),
+    data
+  );
+
+  for (const field of fields) {
+    const name = field.apiName ?? field.name;
+
+    if (isPrimitiveField(field) && name in data) {
+      // Falling back to 'null' as it's what GraphQL users would expect
+      // FIXME: handle locale fallbacks
```
aodhagan-cf, line 33: Would also be nice to capture the fallback in a test

src/live-updates.ts
aodhagan-cf: Btw I think they handle the type in the starter template like this

```
{
    sys: { id: string };
    __typename: string;
    [k: string]: any;
}
```

PR #33, opened 2023-03-23
fix: live updates [TOL-1000, TOL-1022]

YvesRijckaert:
- Primitive fields bug fixes: We only want to update the fields that need to be updated and save the updates in a Map so we check if we already had some changes before entityChanged fires again
- Rrich text GraphQL
- Multi ref field changes (position + removal)

### DIFF
--- a/src/graphql/entries.ts
+++ b/src/graphql/entries.ts
@@ -1,21 +1,23 @@
 import { ContentTypeProps, EntryProps } from 'contentful-management/types';
 
+import { CollectionItem, SysProps } from '../types';
 import { isPrimitiveField, logUnrecognizedFields } from './utils';
 
 /**
  * Updates GraphQL response data based on CMA entry object
  *
- * @param contentType entity
- * @param data the GraphQL response to be updated
- * @param update CMA entry object containing the update
- * @param locale code
+ * @param contentType ContentTypeProps
+ * @param data Record<string, unknown> - The GraphQL response to be updated
+ * @param update EntryProps - CMA entry object containing the update
+ * @param locale string - Locale code
+ * @returns Record<string, unknown> - Updated GraphQL response data
  */
 export function updateEntry(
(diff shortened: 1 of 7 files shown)


YvesRijckaert:
Note: I will add tests / cleanup the code in a follow up

src/graphql/entries.ts
```diff
@@ -1,39 +1,96 @@
...
  *
- * @param contentType entity
- * @param data the GraphQL response to be updated
- * @param update CMA entry object containing the update
- * @param locale code
+ * @param contentType ContentTypeProps
+ * @param data Record<string, unknown> - The GraphQL response to be updated
+ * @param update EntryProps - CMA entry object containing the update
+ * @param locale string - Locale code
+ * @returns Record<string, unknown> - Updated GraphQL response data
```
chrishelgert, line 13:
```suggestion
 * @returns Updated GraphQL response data
```

The return type shouldn't be correct as it is without the `sys` information

src/graphql/entries.ts
```diff
@@ -1,39 +1,96 @@
...
+  data: Record<string, unknown> & { sys: SysProps },
   update: EntryProps,
   locale: string
-): Record<string, unknown> {
+): Record<string, unknown> & { sys: SysProps } {
   const modified = { ...data };
   const { fields } = contentType;
 
-  // Warn about unrecognized fields
   logUnrecognizedFields(
```
chrishelgert, line 24: the log and the clone of data into modified could be after the check if the sys.id matches ;)

src/types.ts
```diff
@@ -13,3 +13,12 @@ export enum TagAttributes {
 export type Entity = Record<string, unknown>;
```
chrishelgert, line 13: We can use this one instead of all the `Record<string, unknown>` ;)
and maybe already merge it together with the SysProps?

PR #36, opened 2023-03-24
feat: complex field support for live preview iframe

aodhagan-cf:
Complex field support for graphql live updates

### DIFF
--- a/src/graphql/__tests__/entries.test.ts
+++ b/src/graphql/__tests__/entries.test.ts
@@ -99,6 +99,7 @@ describe('Update GraphQL Entry', () => {
         },
         reference: null,
       };
+      const testAddingEntryId = '18kDTlnJNnDIJf6PsXE5Mr';
       const sendMessageToEditor = vi.spyOn(Utils, 'sendMessageToEditor').mockReturnThis();
       const update = {
         sys: {
@@ -110,7 +111,7 @@ describe('Update GraphQL Entry', () => {
               sys: {
                 type: 'Link',
                 linkType: 'Entry',
-                id: '18kDTlnJNnDIJf6PsXE5Mr',
+                id: testAddingEntryId,
               },
             },
           },
@@ -136,6 +137,8 @@ describe('Update GraphQL Entry', () => {
         },
         reference: null,
       };
+      const testAddingEntryId = '18kDTlnJNnDIJf6PsXE5Mr';
+      const testContentTypeId = 'testContentType';
       const update = {
         sys: {
           id: entry.sys.id,
@@ -146,7 +149,7 @@ describe('Update GraphQL Entry', () => {
               sys: {
                 type: 'Link',
                 linkType: 'Entry',
-                id: '18kDTlnJNnDIJf6PsXE5Mr',
+                id: testAddingEntryId,
               },
             },
           },
@@ -154,11 +157,11 @@ describe('Update GraphQL Entry', () => {
       } as unknown as EntryProps<KeyValueMap>;
 
       const entityReferenceMap = new Map<string, EntryProps | AssetProps>();
-      entityReferenceMap.set('18kDTlnJNnDIJf6PsXE5Mr', {
+      entityReferenceMap.set(testAddingEntryId, {
         sys: {
           contentType: {
             sys: {
-              id: 'testContentType',
+              id: testContentTypeId,
               type: 'TestContentType',
               linkType: 'Entry',
             },
@@ -173,9 +176,10 @@ describe('Update GraphQL Entry', () => {
           id: entry.sys.id,
         },
         reference: {
+          // content type has been adjusted to have capital letter at the start
           __typename: 'TestContentType',
           sys: {
-            id: '18kDTlnJNnDIJf6PsXE5Mr',
+            id: testAddingEntryId,
             linkType: 'Entry',
             type: 'Link',
           },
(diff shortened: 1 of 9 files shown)


src/graphql/__tests__/entries.test.ts
```diff
@@ -18,12 +19,14 @@ describe('Update GraphQL Entry', () => {
     data,
     update = entry as EntryProps,
     locale = EN,
+    entityReferenceMap = new Map<string, EntryProps | AssetProps>(),
```
YvesRijckaert, line 22: do we maybe want to extract the type of entityReferenceMap and add it to `types.ts`?

YvesRijckaert (approved):
👌

src/graphql/__tests__/entries.test.ts
```diff
@@ -87,6 +90,311 @@ describe('Update GraphQL Entry', () => {
...
+
+      // value has not changed, just sends message back to editor
+      expect(updateFn({ data, update })).toEqual({
+        __typename: 'Page',
+        sys: {
+          id: entry.sys.id,
+        },
+        reference: null,
+      });
+      expect(sendMessageToEditor).toHaveBeenCalled();
```
chrishelgert, line 129: is it relevant that it was called with the correct action?

src/graphql/__tests__/entries.test.ts
```diff
@@ -87,6 +90,311 @@ describe('Update GraphQL Entry', () => {
...
+
+      // value has not changed, just sends message back to editor
+      expect(updateFn({ data, update, entityReferenceMap })).toEqual({
+        __typename: 'Page',
+        sys: {
+          id: entry.sys.id,
+        },
+        reference: {
+          // content type has been adjusted to have capital letter at the start
+          __typename: 'TestContentType',
```
chrishelgert, line 180: it should be the same as in line 165 right?
is the testdata in 164 wrong or the comment or did I get it wrong?  🤯

src/graphql/__tests__/entries.test.ts
```diff
@@ -87,6 +90,311 @@ describe('Update GraphQL Entry', () => {
...
+                linkType: 'Entry',
+                id: testAddingEntryId,
+              },
+            },
+          },
+        },
+      } as unknown as EntryProps<KeyValueMap>;
+
+      // value has not changed, just sends message back to editor
+      expect(updateFn({ data, update })).toEqual({
```
chrishelgert, line 218: should we test that the fn is not called as it is mentioned in the description?

src/graphql/assets.ts
```diff
@@ -47,9 +47,16 @@ const AssetContentType = {
 export function updateAsset(
```
chrishelgert, line 47: generally about the `update` fn, do you think it would make sense to change them from 5 arguments to one object argument?

src/graphql/entries.ts
```diff
@@ -63,34 +67,90 @@ function updateRichTextField(
   locale: string
 ) {
   if (name in modified) {
+    if (!modified[name]) {
+      modified[name] = {};
+    }
     (modified[name] as { json: unknown }).json = update?.fields?.[name]?.[locale] ?? null;
```
chrishelgert, line 73: would we still need the typecast if we define the type it in L71?

src/graphql/entries.ts
```diff
@@ -63,34 +67,90 @@ function updateRichTextField(
...
+  updatedReference: EntryProps & { __typename?: string },
+  entityReferenceMap: Map<string, EntryProps | AssetProps>
+) {
+  // if the reference was deleted return null
+  if (updatedReference === null) {
+    return null;
+  }
+
+  // it's already in graphql format so we can return
+  if (updatedReference && updatedReference.__typename) {
```
chrishelgert, line 101:
```suggestion
  if (updatedReference.__typename) {
```

according to the types it can not be undefined, so this should be enough?

src/live-updates.ts
```diff
@@ -21,17 +23,24 @@ export class LiveUpdates {
     initial: Argument,
     locale: string,
     updated: Entity,
-    contentType: any
+    contentType: any,
+    entityReferenceMap: Map<string, EntryProps | AssetProps>
   ): Argument {
     if ((initial as any).__typename === 'Asset') {
```
chrishelgert, line 29: initial could be an `Array` could you maybe adjust the code 🙏

src/types.ts
```diff
@@ -21,4 +21,11 @@ export interface SysProps {
 
 export interface CollectionItem {
   sys: SysProps;
+  __typename?: string;
+}
+
+export enum MessageAction {
```
chrishelgert, line 27: FYI:
if you define a enum as `const enum` it would be transformed during the build process and the outcome would be smaller & treeshakeable.
Con: you need to iterate over it in a different way

src/live-updates.ts
chrishelgert: ignore this one, I already adjusted it locally

src/graphql/__tests__/entries.test.ts
aodhagan-cf: Good idea! done

src/graphql/__tests__/entries.test.ts
aodhagan-cf: Yeah good idea

src/graphql/entries.ts
aodhagan-cf: unfortunately yes 😢

src/types.ts
aodhagan-cf: Interesting! 🤯

chrishelgert (approved):

PR #46, opened 2023-03-27
docs: add section about gatsby

YvesRijckaert:
Add Gatsby integration documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,13 @@ It uses [Typescript](https://www.typescriptlang.org/), [React](https://reactjs.o
 
 To install live preview simply run one of the following commands.
 
-```
+```bash
 yarn add @contentful/live-preview
 ```
 
 or
 
-```
+```bash
 npm install @contentful/live-preview
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -60,13 +60,98 @@ import '@contentful/live-preview/dist/style.css';
...
+  heading: String!
+  text: String
+}
+```
+
+4. Update the corresponding component to load the contentful_id property:
+
+```jsx
+export const query = graphql`
+  fragment HomepageHeroContent on HomepageHero {
```
chrishelgert, line 110: I needed to adjust also the ContentfulHomepageHeroContent in my example :(

README.md
```diff
@@ -60,13 +60,98 @@ import '@contentful/live-preview/dist/style.css';
...
+import '@contentful/live-preview/dist/style.css';
+import { ContentfulLivePreview } from '@contentful/live-preview';
+
+ContentfulLivePreview.init();
+```
+
+3. In order to tag fields and use live updates, you need to add the contentful_id property to the GraphQL schema. For example, to extend the HomepageHero interface:
+
+```graphql
+interface HomepageHero {
```
chrishelgert, line 98: it this the part in the gatsby-node.js?
Then I had to also adjust this one: `ContentfulHomepageHero` :(

YvesRijckaert: Ah yes I will add it as well 👌

PR #47, opened 2023-03-28
fix: merging graphql properties that we already have [TOL-1079]

aodhagan-cf:
When resorting we can get an issue where graphql properties are lost. This makes sure that if we have them we merge them with the updated order etc.

### DIFF
--- a/src/graphql/entries.ts
+++ b/src/graphql/entries.ts
@@ -103,7 +103,7 @@ function updateReferenceField(
   );
   // if we have the typename of the updated reference, we can return with it
   if (entityTypename) {
-    return { ...updatedReference, __typename: entityTypename, ...referenceFromPreviewApp };
+    return { ...referenceFromPreviewApp, ...updatedReference, __typename: entityTypename, };
   } else {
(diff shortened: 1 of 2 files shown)


src/graphql/entries.ts
```diff
@@ -98,7 +103,7 @@ function updateReferenceField(
   );
   // if we have the typename of the updated reference, we can return with it
   if (entityTypename) {
-    return { ...updatedReference, __typename: entityTypename };
+    return { ...updatedReference, __typename: entityTypename, ...referenceFromPreviewApp };
```
chrishelgert, line 106:
```suggestion
    return { ...referenceFromPreviewApp, ...updatedReference, __typename: entityTypename, };
```

shouldn't it be like that? otherwise the `referenceFromPreviewApp` could overwrite updatedReference again 🤔

aodhagan-cf: can write a test

aodhagan-cf: @chrishelgert have added tests for single and multi ref to keep graphql properties

chrishelgert (approved):
Thanks 🤝

PR #48, opened 2023-03-29
chore(live-updates): prepare restore changes after navigation [TOL-1071]

chrishelgert:
Initial implementation of the restoration of updates after navigation (currently disabled) and some cleanups

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ import '@contentful/live-preview/style.css';
 
 ### Live Updates
 
-Live Updates from the editor to your applications are currently only directed supported for [React.js](https://reactjs.org/)
+Live Updates from the editor to your applications are out of the box only supported for [React.js](https://reactjs.org/) at the moment.
 
 ```tsx
 import { useContentfulLiveUpdates } from "@contentful/live-preview/react";
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@ It uses [Typescript](https://www.typescriptlang.org/), [React](https://reactjs.o
 ### Requirements
 
 - Node.js: `>=16.15.1`
-- Yarn: `>=1.21.1`
 
 To install live preview simply run one of the following commands.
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -57,7 +56,7 @@ import '@contentful/live-preview/dist/style.css';
 
 ### Live Updates
 
-Live Updates from the editor to your applications are currently only supported for [React.js](https://reactjs.org/)
+Live Updates from the editor to your applications are currently only directed supported for [React.js](https://reactjs.org/)
```
YvesRijckaert, line 59:
```suggestion
Live Updates from the editor to your applications are currently only directly supported for [React.js](https://reactjs.org/)
```

YvesRijckaert: Or maybe 'out of the box support only for react'?

2wce (approved):

YvesRijckaert (approved):

src/types.ts
```diff
@@ -12,6 +14,7 @@ export enum TagAttributes {
   LOCALE = 'data-contentful-locale',
 }
 
+// TODO: can we add sys and optional typename to the Entity?
 export type Entity = Record<string, unknown>;
```
aodhagan-cf, line 18: Should we have 2 types? `GraphqlEntity` & `Entity` in this case

src/utils.ts
```diff
@@ -1,13 +1,28 @@
 import type { MessageAction } from './types';
 
+type IframeConnectedMessage = {
```
aodhagan-cf, line 3: Do you think it would be nice to merge these types with the `MessageAction` enum? It might be annoying maintaining both. Also in future it could be nice to import this type in the web app

aodhagan-cf (approved):

src/types.ts
chrishelgert: GraphQL entity would be the one with __typename, but also the REST one would need the `sys` information, so I'm still not sure whats the best way to define/solve this 💭

chrishelgert: Maybe we gain more insights with the REST implementation 👍

PR #54, opened 2023-03-31
feat: Live Updates REST endpoint support

YvesRijckaert:
- Add support for Live Updates REST endpoint
- Rename variables in GraphQL and LiveUpdates class to make it more descriptive

### DIFF
--- a/src/rest/entries.ts
+++ b/src/rest/entries.ts
@@ -1,14 +1,25 @@
 import { EntryProps, ContentTypeProps } from 'contentful-management/types';
+import { ContentFields } from 'contentful-management/types';
 
 import { SysProps, Entity } from '../types';
+import { isPrimitiveField, updatePrimitiveField } from '../utils';
 
-const mergeFields = (initialObj: any, updatedObj: any, locale: string) => {
+const mergeFields = (fields: ContentFields[], initialObj: any, updatedObj: any, locale: string) => {
   const mergedObj = { ...initialObj };
 
   console.log({ initialObj, updatedObj });
 
-  for (const field in updatedObj.fields) {
-    mergedObj.fields[field] = updatedObj.fields[field][locale];
+  for (const field of fields) {
+    const name = field.apiName ?? field.name;
+
+    if (isPrimitiveField(field) || field.type === 'RichText') {
+      updatePrimitiveField(initialObj.fields, updatedObj, name, locale);
+    } else if (field.type === 'Link') {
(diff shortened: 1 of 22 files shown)


src/rest/entries.ts
```diff
@@ -0,0 +1,72 @@
...
+  const mergedObj = { ...dataFromPreviewApp };
+
+  for (const field of fields) {
+    const name = field.apiName ?? field.name;
+
+    if (isPrimitiveField(field) || field.type === 'RichText') {
+      updatePrimitiveField(dataFromPreviewApp.fields, updateFromEntryEditor, name, locale);
+      mergedObj.fields[name] = dataFromPreviewApp.fields[name];
+    } else if (field.type === 'Link') {
+      // update single ref field
```
aodhagan-cf, line 22: Super nit picky but maybe have these as TODOs

src/utils.ts
aodhagan-cf, line 1: I'm fine with this for now, just worried that this util file will grow, maybe in the future we can split this up a little bit.

```
import { isPrimitiveField, updatePrimitiveField } from './helpers/field'
import { generateUID } from './helpers/uuid'
import { debounce } from './helpers/util'
```

as a suggestion

src/rest/entries.ts
```diff
@@ -0,0 +1,72 @@
...
+  locale: string
+): Entity & { sys: SysProps } {
+  if (dataFromPreviewApp.fields) {
+    const updatedFields = { ...dataFromPreviewApp.fields };
+
+    for (const [fieldName, field] of Object.entries(dataFromPreviewApp.fields)) {
+      if (field.sys && field.sys.id === updateFromEntryEditor.sys.id) {
+        updatedFields[fieldName] = mergeFields(fields, field, updateFromEntryEditor, locale);
+        return { ...dataFromPreviewApp, fields: updatedFields };
+      } else if (typeof field === 'object') {
```
aodhagan-cf, line 44: Is there any other way to know whether we should call this `updateNestedRef`? just a little confused about what is going on here.

YvesRijckaert: It's to update the fields inside of references, if they have been resolved. It's recursive and will try to find the id of the reference and compare it to the id of the update that is coming from the webapp and then updating it

chrishelgert: I'm probably going to change this a little bit anyways, as I kind of need it also for the asset updates - sorry!

PR #59, opened 2023-04-06
feat(live-preview): support for assets [TOL-1055]

chrishelgert:


src/live-updates.ts
```diff
@@ -95,19 +98,95 @@ export class LiveUpdates {
   // TODO: we currently expect the provided data to have the sys.id directly, but it could be also nested
   // { items: [{ sys: { id }, ... }] }
   // do we support this case or should it be provided like that: [{ sys: { id } }]
-  private mergeEntity(data: MergeEntityProps): Entity {
-    if ('__typename' in data.dataFromPreviewApp) {
-      return this.mergeGraphQL(data);
+  // once there is a match, update the match with either the GQL or REST fn
+  // only call the `cb` if there was an update
+  private mergeNestedReference(
```
chrishelgert, line 103: this needs cleanup 🤯

src/rest/entries.ts
```diff
@@ -26,47 +34,4 @@ const mergeFields = (
   }
 
   return mergedObj;
-};
-
-function updateNestedRef(
```
chrishelgert, line 31: Is now part of the live-updates so we've to maintain it only once

src/graphql/assets.ts
```diff
@@ -44,19 +10,29 @@ const AssetContentType = {
  * @param asset CMA Asset object containing the update
  * @param locale locale code
  */
-export function updateAsset(
-  data: Entity & { sys: SysProps },
+export function updateAsset<T extends (Entity & { sys: SysProps }) | EntryProps>(
+  data: T,
```
YvesRijckaert, line 14:
```suggestion
  dataFromPreviewApp: T,
```

src/graphql/assets.ts
```diff
@@ -44,19 +10,29 @@ const AssetContentType = {
...
   update: AssetProps,
```
YvesRijckaert, line 15:
```suggestion
  updateFromEntryEditor: AssetProps,
```

src/graphql/assets.ts
YvesRijckaert: I find the naming here a bit more clear, it's also in line with the other naming in the other files

YvesRijckaert (approved):
Looking good!

src/graphql/assets.ts
chrishelgert: yes you're right, oversaw that it has still the old names 🙈

package.json
```diff
@@ -66,6 +65,7 @@
     "@types/react-dom": "^18.0.10",
     "@vitejs/plugin-react-swc": "^3.0.0",
     "commitizen": "4.3.0",
+    "contentful-management": "^10.30.1",
```
aodhagan-cf, line 68: just a dev dep because we just use the types right?

package.json
```diff
@@ -52,7 +52,6 @@
   },
   "homepage": "https://github.com/contentful/live-preview#readme",
   "dependencies": {
-    "contentful-management": "^10.30.1",
```
chrishelgert, line 55: We only need the types, no need for customers to install it :)

src/graphql/__tests__/asset.test.ts
```diff
@@ -0,0 +1,7 @@
+import { describe, it } from 'vitest';
+
+describe('Update GraphQL Assets', () => {
+  it('merges the data together and resolves the nested information', () => {
```
aodhagan-cf, line 4: Maybe a skip on the non-implemented tests?

aodhagan-cf: actually if you are doing them tomorrow no worries

src/graphql/assets.ts
```diff
@@ -1,62 +1,37 @@
...
+      ...dataFromPreviewApp,
+      // GraphQL flattens some information
+      // and as the live updates are coming from the CMA, we need to transform them
+      title: updateFromEntryEditor.fields.title[locale],
+      description: updateFromEntryEditor.fields.description?.[locale],
+      contentType: file.contentType,
+      width: file.details?.image.width,
+      height: file.details?.image.height,
+      // GraphQL returns the URL with protocol and the CMA without it, so we need to add the information in there manually
+      url: file.url ? (file.url?.startsWith('https:') ? file.url : `https:${file.url}`) : undefined,
```
aodhagan-cf, line 31: Could maybe extract this to a util function

```
function updateProtocolToHttps(urlString: string | undefined): string | undefined {
  if (typeof urlString !== 'string') {
    return urlString;
  }

  try {
    const url = new URL(urlString);
    url.protocol = 'https:';
    return url.href;
  } catch (error) {
    console.error(`Error updating protocol to https: ${error}`);
    return urlString;
  }
}
```

src/graphql/entries.ts
```diff
@@ -83,19 +85,76 @@ function getContentTypenameFromEntityReferenceMap(
...
     if (entity) {
-      const contentTypeId = entity.sys.contentType?.sys.id;
+      const contentTypeId = entity.sys.contentType.sys.id;
       const typename = contentTypeId.charAt(0).toUpperCase() + contentTypeId.slice(1);
       return typename;
     }
   }
 }
 
+const ASSET_TYPENAME = 'Asset';
```
aodhagan-cf, line 95: This to me feels like it could live in a `graphql/types.ts` file

src/live-updates.ts
```diff
@@ -33,81 +43,129 @@ export class LiveUpdates {
...
-      //@ts-expect-error -- ..
-      cachedData,
-      updateFromEntryEditor,
-      locale,
-      entityReferenceMap
-    );
-
-    // Cache the updated data for future updates
-    this.storage.set(entryId, updatedData);
+    if (this.isEntryOrAsset(dataFromPreviewApp)) {
```
aodhagan-cf, line 77: `isEntity`?

src/live-updates.ts
```diff
@@ -33,81 +43,129 @@ export class LiveUpdates {
...
+      // REST
+      return {
+        data: rest.updateEntry(
```
aodhagan-cf, line 80: should this be `updateEntity`?

src/graphql/assets.ts
chrishelgert: Needed to adjust it, as `new URL('//images.ctfassets.net/')` is invalid, it needs the protocol :(

src/graphql/entries.ts
chrishelgert: added it behind a debug flat, that the customer can enable 👍

src/helpers/utils.ts
```diff
@@ -82,3 +83,42 @@ export class StorageMap<T extends unknown> {
...
+ * Hint: It uses the structuredClone which is only available in modern browsers,
+ * for older one it uses the JSON.parse(JSON.strinfify) hack.
+ */
+export function clone<T extends Record<string, unknown> | Array<unknown>>(incoming: T): T {
+  if (typeof structuredClone === 'function') {
+    return structuredClone(incoming);
+  }
+
+  try {
+    return JSON.parse(JSON.stringify(incoming));
```
aodhagan-cf, line 119: I think this will break if we have a Map in the object, you okay with that?

aodhagan-cf: I think it's fine for our update use case, just calling it out

src/helpers/utils.ts
```diff
@@ -82,3 +83,42 @@ export class StorageMap<T extends unknown> {
...
+ */
+export function setProtocolToHttps(url: string | undefined): string | undefined {
+  if (!url) {
+    return url;
+  }
+
+  try {
+    // new URL('//images.ctfassets.net') is invalid, therefore we add it with protocol as `base`.
+    // This will merge it then correct together
+    const parsed = new URL(url, 'https://images.ctfassets.net');
```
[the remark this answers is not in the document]
chrishelgert: If you use `new URL` you can define a base domain as second argument (necessary for relative paths).
If the provided url starts with `//` the function will think it is relative and uses the base and then merges them together.
Without the `base domain`, new URL is not able to handle `//images.ctfassets.net` :(

src/helpers/utils.ts
chrishelgert: Yeah it's already the fallback, modern browsers (that support map) should also have this `structuredClone` function which is a native deep merge fn.
If that should not work, we try it with `JSON.parse(JSON.stringify)` thing, and otherwise we will work with the incoming one as laaaast fallback

src/types.ts
```diff
@@ -14,16 +15,29 @@ export const enum TagAttributes {
...
 }
 
+export type Entity = Record<string, unknown>;
+export interface EntityWithSys extends Entity {
+  sys: SysProps;
+  __typename?: string;
+}
+
+export function hasSysInformation(entity: unknown): entity is EntityWithSys {
+  return !!(
```
aodhagan-cf, line 30: `entity && typeof entity === 'object' && entity.satisfies((e) => 'sys' in e && e.sys.id);`

Just showing, no need to change

aodhagan-cf: I wonder would `!!entity?.satisfies((e) => 'sys' in e && e.sys.id)` even work?

src/rest/entries.ts
```diff
@@ -58,15 +24,23 @@ function updateNestedRef(
...
-  } else {
-    return updateNestedRef(contentType.fields, dataFromPreviewApp, updateFromEntryEditor, locale);
+): EntryProps {
+  for (const field of contentType.fields) {
+    const name = getFieldName(contentType, field);
+
+    if (isPrimitiveField(field) || field.type === 'RichText' || field.type === 'File') {
+      updatePrimitiveField(dataFromPreviewApp.fields, updateFromEntryEditor, name, locale);
+    } else if (field.type === 'Link') {
+      // update single ref field
```
aodhagan-cf, line 39: What's happening here? are these TODOs?

chrishelgert: yes it's this story: https://contentful.atlassian.net/browse/TOL-1025
(Was already this comment before, the position has only changed)

src/live-updates.ts
```diff
@@ -33,81 +43,129 @@ export class LiveUpdates {
...
+   * caches the result if cache is enabled and the entity has a `sys.id`.
+   * Caching should not be enabled for every entry,
+   * because nested references could be merged differently together and this could solve to data loss.
+   */
+  private mergeNestedReference(
+    { dataFromPreviewApp, ...params }: MergeEntityProps,
+    useCache: boolean
+  ): { data: Entity; updated: boolean } {
+    const dataFromPreviewappId = hasSysInformation(dataFromPreviewApp) && dataFromPreviewApp.sys.id;
+    const isCacheable = useCache && dataFromPreviewappId;
```
aodhagan-cf, line 105: nit: what if the sys id is 0?

chrishelgert: Our ID's are always strings :)

aodhagan-cf (approved):

PR #74, opened 2023-04-17
feat: notifiy the editor about subscriptions [TOL-1059]

chrishelgert:
Send a message to the editor once there is a subscription and if it is using GQL or REST

### DIFF
--- a/src/liveUpdates.ts
+++ b/src/liveUpdates.ts
@@ -1,7 +1,7 @@
 import type { AssetProps, EntryProps } from 'contentful-management';
 
 import * as gql from './graphql';
-import { clone, generateUID, StorageMap } from './helpers';
+import { clone, generateUID, sendMessageToEditor, StorageMap, debug } from './helpers';
 import * as rest from './rest';
 import {
   Argument,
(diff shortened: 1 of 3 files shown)


aodhagan-cf (approved):
I guess another way of identifying is if it have the fields property right?

chrishelgert:
> I guess another way of identifying if it's REST is if it have the fields property right?

yes, should we do it to make sure? 🤔

src/liveUpdates.ts
```diff
@@ -210,16 +210,66 @@ export class LiveUpdates {
...
+
+    if (!hasSys) {
+      isValid = false;
+      debug.error('Live Updates requires the "sys.id" to be present on the provided data', data);
+    }
+
+    if (!isGQL && !isREST) {
+      isValid = false;
+      debug.error(
+        'Live Updates requires the the provided data to contain either "fields" or "__typename"',
```
aodhagan-cf, line 235: I think we can be explicit here for REST and Graphql.

`For live updates as a basic requirement the provided data must include the "fields" property for REST or "__typename" for Graphql, otherwise the data will be treated as invalid and live updates will not work.`

what do you think about something like this?

PR #75, opened 2023-04-18
feat(rest): add support for references [TOL-1025]

chrishelgert:
Adding support for References in the REST

- [x] Single Reference
- [x] Multi Reference
- [x] Adding tests


src/helpers/field.ts
```diff
@@ -6,6 +6,7 @@ const PRIMITIVE_FIELDS = new Set([
   'Symbol',
   'Text',
   'Integer',
+  'Number',
```
chrishelgert, line 9: Decimal was not supported yet 🙈

src/rest/__tests__/fixtures/dataFromPreviewApp.json
```diff
@@ -18,7 +18,7 @@
     "shortTextUrl": "https://example.com",
     "shortTextSlug": "all-fields-page",
     "shortTextList": ["short", "text", "list", "comma"],
-    "longText": "long\ntext\nmultiple",
+    "longTextMultiple": "long\ntext\nmultiple",
```
chrishelgert, line 21: Was also not working in the test, as there was no match with the content type yet

src/rest/__tests__/fixtures/updateFromEntryEditor.json
```diff
@@ -18,10 +18,10 @@
   },
   "fields": {
     "shortText": { "en-US": "test test test test test " },
-    "shortTextUrl": { "en-US": "https://example.com" },
-    "shortTextSlug": { "en-US": "all-fields-page" },
+    "shortTextUrl": { "en-US": "https://app.contentful.com" },
```
chrishelgert, line 21: Updated more things, to test all primitive fields

src/styles.css
```diff
@@ -1,9 +1,9 @@
-[data-contentful-field-id] {
+[data-contentful-field-id][data-contentful-entry-id][data-contentful-locale] {
```
chrishelgert, line 1: Show only the borders if everything is configured correctly (caused confusion, because the border was shown, but no edit button)

src/types.ts
```diff
@@ -83,4 +83,4 @@ export type MessageFromSDK = EditorMessage & {
   location: string;
 };
 
-export class EntryReferenceMap extends Map<string, EntryProps | AssetProps> {}
+export class EntityReferenceMap extends Map<string, EntryProps | AssetProps> {}
```
chrishelgert, line 86: renamed as it contains assets and entries

PR #76, opened 2023-04-19
feat: keep info after navigating [TOL-1080]

YvesRijckaert:
- Activate persisting data across page navigation
- Ensure callback is only called for active subscriptions in restore method
- Make restore method work with Arrays

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -40,20 +40,13 @@ export function debounce<T extends Callback>(func: T, timeout = 100): DebouncedF
 export class StorageMap<T extends unknown> {
   private storageKey: string;
   private value: Map<string, T>;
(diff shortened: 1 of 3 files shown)


src/helpers/utils.ts
```diff
@@ -70,16 +63,23 @@ export class StorageMap<T extends unknown> {
...
-        // ignored
-      }
+    try {
+      // Attention: Map can not be `JSON.stringify`ed directly
+      window.sessionStorage.setItem(
+        this.storageKey,
+        JSON.stringify(Array.from(this.value.entries()))
+      );
+    } catch (err) {
+      // ignored
```
chrishelgert, line 73: would you mind replacing the `// ignored` through our internal `debug.warn`?

chrishelgert (approved):

PR #90, opened 2023-05-03
chore(docs): remove alpha and add official docs

chrishelgert:
- remove the alpha state note
- add a link to the official developer documentation

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,7 +7,7 @@ import { Argument, LivePreviewProps, SubscribeCallback, TagAttributes } from './
 
 interface ContentfulLivePreviewInitConfig {
   debugMode?: boolean;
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -7,7 +7,7 @@ import { Argument, LivePreviewProps, SubscribeCallback, TagAttributes } from './
 
 interface ContentfulLivePreviewInitConfig {
   debugMode?: boolean;
-  enableFieldTagging?: boolean;
+  enableInspectorMode?: boolean;
```
z0al, line 10: This is a breaking change. Is it intentional?

chrishelgert: yes (was only brought in recently so low risk)

z0al (approved):

PR #104, opened 2023-05-08
feat: add locale in init [TOL-1152]

2wce:
This should optimise the integration with the SDK & only ask the client to add the locale once


YvesRijckaert (changes requested):
Can we update the readme? Also maybe add a section about overwriting the locale?

chrishelgert (commented):
Sorry caused you some merge conflicts 😿 
As the locale is now required in the `init` is it a breaking change?
Otherwise looks good 👍

2wce:
@chrishelgert I've resolved the merge conflicts 👍🏼

src/styles.css
```diff
@@ -3,7 +3,7 @@
   transition: outline-color 0.3s ease-in-out;
 }
 
-.contentful-inspector--active [data-contentful-field-id][data-contentful-entry-id][data-contentful-locale] {
+.contentful-inspector--active [data-contentful-field-id][data-contentful-entry-id] {
```
chrishelgert, line 6: we should remove the locale also from line 1, otherwise the animation should not work

src/index.ts
```diff
@@ -128,15 +145,15 @@ export class ContentfulLivePreview {
...
+    if (!fieldId || !entryId) {
+      debug.warn('Missing property for inspector mode', { fieldId, entryId });
       return null;
     }
 
     return {
       [TagAttributes.FIELD_ID]: fieldId,
       [TagAttributes.ENTRY_ID]: entryId,
-      [TagAttributes.LOCALE]: locale,
+      [TagAttributes.LOCALE]: locale ? locale : this.locale,
```
chrishelgert, line 156: we would only need to render it here if a custom locale is provided, the inspectorMode class could get the locale from the init function and use it as fallback if the tag is not there 🤔

src/react.tsx
```diff
@@ -86,10 +90,15 @@ export function useContentfulLiveUpdates<T extends Argument | null | undefined>(
...
     }
     // or update content through live updates
-    return ContentfulLivePreview.subscribe(data as Argument, locale, (updatedData) => {
-      // Update the state and adding a version number to it, as some deep nested updates
-      // are not proceeded correctly otherwise
-      update.current((prevState) => ({ data: updatedData as T, version: prevState.version++ }));
+    return ContentfulLivePreview.subscribe({
+      data: data as Argument,
+      // override locale with user locale if passed otherwise use the local from the initialisation
+      locale: locale ? locale : ContentfulLivePreview.locale,
```
chrishelgert, line 96: the locale handling is also in the in the subscibe fn (https://github.com/contentful/live-preview/pull/104/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R135)
could we remove it in the react part?

src/types.ts
```diff
@@ -4,9 +4,9 @@ export type ContentType = ContentTypeProps;
 export const ASSET_TYPENAME = 'Asset';
 
 export type LivePreviewProps = {
-  fieldId: string;
-  entryId: string;
-  locale: string;
+  fieldId: string | null | undefined;
```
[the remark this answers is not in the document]
2wce: I'm not sure, I think those changes were in an earlier commit

src/index.ts
2wce: do you mean at this point?  https://github.com/contentful/live-preview/blob/155a1483b19b92e35b551d8530b9381901c701ca/src/inspectorMode.ts#L73

chrishelgert: yes 👍

chrishelgert (approved):
LGTM

YvesRijckaert (approved):

PR #106, opened 2023-05-12
fix(live-updates): add support for videos, currently they are not processed correctly [TOL-1173]

chrishelgert:
Videos doesn't have the width and height information, so therefor they are currently not updating correctly.
This can lead to issues on the app side as the returned information are not the same as expected.


YvesRijckaert (approved):
📺

aodhagan-cf (approved):

PR #114, opened 2023-05-15
fix: conditionally check correct props [NONE]

2wce:
We had a bug with the inspector props & the locale was treated as a required prop for inspector mode even after the locale optimisation was added. this should fix that

### DIFF
--- a/src/react.tsx
+++ b/src/react.tsx
@@ -112,7 +112,7 @@ type GetInspectorModeProps<T> = (props: {
  * Generates the function to build the required properties for the inspector mode (field tagging)
  */
 export function useContentfulInspectorMode<
-  T = undefined | Pick<LivePreviewProps, 'entryId'> | Pick<LivePreviewProps, 'entryId' | 'locale'>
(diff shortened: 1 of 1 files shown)


YvesRijckaert (approved):

src/react.tsx
```diff
@@ -112,7 +112,7 @@ type GetInspectorModeProps<T> = (props: {
  * Generates the function to build the required properties for the inspector mode (field tagging)
  */
 export function useContentfulInspectorMode<
-  T = undefined | Pick<LivePreviewProps, 'entryId'> | Pick<LivePreviewProps, 'entryId' | 'locale'>
+  T = undefined | Pick<LivePreviewProps, 'entryId'> | Pick<LivePreviewProps, 'entryId' | 'fieldId'>
```
chrishelgert, line 115: Probably we could also remove the last block `Pick<LivePreviewProps, 'entryId' | 'fieldId'>` because the function will not be re-useable anymore? or we would need to adjust the typing for the `GetInspectorModeProps` because currently you would still need to provide an object as param 🤔

PR #112, opened 2023-05-15
fix(graphql): support for adding references inside nested references [TOL-1181]

chrishelgert:
- add functionality for nested reference updates
- more explained error messages for required params on init and provider
- add more tests / update locale handling in tests
- typing improvments
- remove false positives for unrecognized fields

Tickets:
- https://contentful.atlassian.net/browse/TOL-1181
- https://contentful.atlassian.net/browse/TOL-1172

PR #117, opened 2023-05-16
chore: exclude locale from type mapping [NONE}

2wce:
The type mapping for GetInspectorModeProps was resulting in the locale being set to required when it shouldn't be

### DIFF
--- a/src/react.tsx
+++ b/src/react.tsx
@@ -104,9 +104,14 @@ export function useContentfulLiveUpdates<T extends Argument | null | undefined>(
   return state.data;
 }
 
-type GetInspectorModeProps<T> = (props: {
-  [K in Exclude<keyof LivePreviewProps, keyof T>]: LivePreviewProps[K];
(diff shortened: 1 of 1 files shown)


chrishelgert:
Hey,
found the same yesterday in https://github.com/contentful/live-preview/pull/112/files#diff-5d3965220b4d780354e048e353b782b560d2328ea4fc413d4bdca448884cf716R107 :D

any preference for the type?

2wce:
nice 👌🏽 no preference, we can close this & use the solution in the draft PR

chrishelgert:
> > nice 👌🏽 no preference, we can close this & use the solution in the draft PR
> 
> Is there anything we can do to get that PR ready for review then? Would be nice to unblock this type issue

ok to unblock you we could also ship this one first, but we have to merge it as `fix` commit otherwise it will not trigger a new version

chrishelgert (approved):
Please merge it with a `fix:` commit message

PR #118, opened 2023-05-17
feat: next js example

YvesRijckaert:


examples/nextjs/.env.local.example
```diff
@@ -0,0 +1,4 @@
+CONTENTFUL_SPACE_ID=
+CONTENTFUL_ACCESS_TOKEN=
+CONTENTFUL_PREVIEW_ACCESS_TOKEN=
+CONTENTFUL_PREVIEW_SECRET=
```
MayaGillilanContentful, line 4: Maybe let's add some comment explaining what each of these are, like in the templates .env.example files

examples/nextjs/README.md
```diff
@@ -0,0 +1,43 @@
+# Next.js Contentful live preview SDK example
+
+This is an example project that demonstrates how to use the `@contentful/live-preview` SDK with a Next.js application. The SDK provides live preview functionality for content changes and the inspector mode for your Contentful space.
+
+## 1. Installation
+
+Install the dependencies:
+
+```bash
+npm install
```
MayaGillilanContentful, line 10: npm or yarn?

examples/nextjs/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+In order to enable the live preview feature in your local development environment, you need to set up the Content preview URL in your Contentful space.
+
+`http://localhost:3000/api/preview?secret=<CONTENTFUL_PREVIEW_SECRET>&slug={entry.fields.slug}`
+
+Replace `<CONTENTFUL_PREVIEW_SECRET>` with its respective value in `.env.local`.
+
+## 5. Running the project locally
+
+To run the project locally, you can use the `npm run dev` command. You can now use the live preview feature.
```
MayaGillilanContentful, line 43: also npm or yarn?

examples/nextjs/next.config.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  async headers() {
+    return [
+      {
+        source: '/(.*)',
+        headers: [
+          {
+            key: 'X-Frame-Options',
+            value: 'SAMEORIGIN https://app.contentful.com',
+          },
```
MayaGillilanContentful, line 10:
```suggestion
          {
            key: 'X-Frame-Options',
            value: 'SAMEORIGIN',
          },
          {
            key: 'Content-Security-Policy',
            value: `frame-ancestors 'self' https://app.contentful.com`,
          },
```
This is deprecated, we should be doing it this way now

examples/nextjs/pages/_app.js
```diff
@@ -0,0 +1,12 @@
+import { ContentfulLivePreviewProvider } from "@contentful/live-preview/react";
+import "@contentful/live-preview/style.css";
+
+function App({ Component, pageProps }) {
+  return (
+    <ContentfulLivePreviewProvider locale="en-US" debugMode={true}>
```
MayaGillilanContentful, line 6: Should we also pass these in?
```
      enableInspectorMode={pageProps.previewActive}
      enableLiveUpdates={pageProps.previewActive}
```

examples/nextjs/next.config.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  async headers() {
+    return [
+      {
+        source: '/(.*)',
```
chrishelgert, line 5:
```suggestion
        source: '/:path*',
```

otherwise it will not be applied on the homepage

MayaGillilanContentful (changes requested):

examples/nextjs/pages/posts/[slug].js
```diff
@@ -0,0 +1,50 @@
...
+  const updatedPost = useContentfulLiveUpdates(post);
+  const inspectorProps = useContentfulInspectorMode({ entryId: post.sys.id });
+
+  if (!router.isFallback && !updatedPost) {
+    return <ErrorPage statusCode={404} />;
+  }
+
+  return (
+    <>
+      {updatedPost.title && (
```
Comment (chrishelgert) [inline examples/nextjs/pages/posts/[slug].js:21]: we could also render the `<h1>` but fallback to an empty string (if it is required in the contentType)

examples/nextjs/README.md
YvesRijckaert: We have a package-lock file so it's recommended to use npm :)

examples/nextjs/README.md
YvesRijckaert: see above ^

examples/nextjs/next.config.js
aodhagan-cf, line 1: Did none of the other headers in the marketing template seem relevant?

examples/nextjs/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "private": true,
+  "scripts": {
+    "dev": "next",
+    "build": "next build",
+    "start": "next start"
+  },
+  "dependencies": {
+    "@contentful/live-preview": "^2.0.2",
+    "next": "latest",
```
aodhagan-cf, line 10: Is this always guaranteed to work with latest? might be safer to select a version just in case there are any breaking changes in next

examples/nextjs/README.md
MayaGillilanContentful: Ahh ok, I missed that. Was a bit confused since we use yarn for the main package

examples/nextjs/next.config.js
YvesRijckaert: I'd keep this short and sweet, only the CSP headers are relevant to the iframe embedding. And this example only shows relevant code to live preview

aodhagan-cf: Okay cool, makes sense

MayaGillilanContentful (approved):

PR #123, opened 2023-05-22
fix(references): move to promises to optimize loading behaviour

chrishelgert:
1. We have the case that not everything is resolved correctly on nested references
2. We would need to return only partial data to resolve then in the next step the reference correctly

This two cases can lead to a broken UI if the handling of data is not checking things very precise on their part.
Let's make the entity not known a promise and build everything async.

- [ ] Update message sender to include the action before merging this one

### DIFF
--- a/src/__tests__/resolveReference.spec.ts
+++ b/src/__tests__/resolveReference.spec.ts
@@ -57,7 +57,7 @@ describe('resolveReference', () => {
     });
 
     it('resolves it from the editor', async () => {
-      const prom = resolveReference({
+      const promise = resolveReference({
         entityReferenceMap: new Map(),
         referenceId: '1',
         isAsset: true,
@@ -79,13 +79,13 @@ describe('resolveReference', () => {
         },
       } as MessageEvent);
 
-      const result = await prom;
+      const result = await promise;
 
       expect(result).toEqual(expected);
     });
 
     it('ignores different entries while waiting for the reference', async () => {
-      const prom = resolveReference({
+      const promise = resolveReference({
         entityReferenceMap: new Map(),
         referenceId: '1',
         isAsset: true,
@@ -123,19 +123,19 @@ describe('resolveReference', () => {
         },
       } as MessageEvent);
 
-      const result = await prom;
+      const result = await promise;
 
       expect(result).toEqual(expected);
     });
 
     it('dedupes requests', async () => {
-      const prom = resolveReference({
+      const promise = resolveReference({
         entityReferenceMap: new Map(),
         referenceId: '1',
         isAsset: true,
       });
 
-      const prom2 = resolveReference({
+      const promise2 = resolveReference({
         entityReferenceMap: new Map(),
         referenceId: '1',
         isAsset: true,
@@ -149,8 +149,8 @@ describe('resolveReference', () => {
         },
       } as MessageEvent);
 
-      const result = await prom;
-      const result2 = await prom2;
+      const result = await promise;
+      const result2 = await promise2;
 
       expect(result).toEqual(expected);
       expect(result2).toEqual(expected);
@@ -198,7 +198,7 @@ describe('resolveReference', () => {
     });
 
     it('resolves it from the editor', async () => {
-      const prom = resolveReference({
+      const promise = resolveReference({
         entityReferenceMap: new Map(),
         referenceId: '1',
       });
@@ -219,7 +219,7 @@ describe('resolveReference', () => {
         },
       } as MessageEvent);
 
-      const result = await prom;
+      const result = await promise;
 
       expect(result).toEqual(expected);
     });
--- a/src/__tests__/resolveReference.spec.ts
+++ b/src/__tests__/resolveReference.spec.ts
@@ -1,0 +1,227 @@
+/* @vitest-environment jsdom */
+import type { AssetProps, EntryProps } from 'contentful-management';
+import { beforeEach, describe, vi, it, afterEach, expect } from 'vitest';
+
+import { sendMessageToEditor } from '../helpers';
+import { resolveReference } from '../helpers/resolveReference';
+import { ASSET_TYPENAME } from '../types';
+
+vi.mock('../helpers/utils');
+
+describe('resolveReference', () => {
+  const listener = vi.spyOn(window, 'addEventListener');
+
+  beforeEach(() => {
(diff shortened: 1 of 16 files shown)


src/graphql/entries.ts
```diff
@@ -109,51 +111,83 @@ function isAsset(entity: EntryProps | (Entity & CollectionItem)): boolean {
...
   return updateAsset(
     { ...referenceFromPreviewApp, ...updatedReference, __typename: ASSET_TYPENAME },
-    match,
+    match as AssetProps,
     locale
   );
 }
 
-function updateReferenceEntryField(
+async function getReference(
```
chrishelgert, line 132: This is the relevant part, currently it leads to overfetching, we need to extract this into a singleton to load things

src/types.ts
```diff
@@ -68,6 +68,12 @@ type UnknownEntityMessage = {
   referenceContentType?: string;
 };
 
+type UnknownEntityPromiseMessage = {
```
chrishelgert, line 71: Combine with the original one and add a version on it

src/graphql/entries.ts
```diff
@@ -109,51 +111,83 @@ function isAsset(entity: EntryProps | (Entity & CollectionItem)): boolean {
...
-    // where we can calculate the asset on the next update message.
-    sendMessageToEditor({
-      action: 'ENTITY_NOT_KNOWN',
-      referenceEntityId: updatedReference.sys.id,
-      referenceContentType: ASSET_TYPENAME,
-    });
-    return;
-  }
+  console.log('>> update asset field', { referenceFromPreviewApp, updatedReference });
+  const { match } = await getReference(entityReferenceMap, updatedReference.sys.id, true);
```
aodhagan-cf, line 121: Being super picky here but maybe makes sense for this to be an object? For example I have no idea what `true` is here at first glance

aodhagan-cf: That or creating a separate function for assets

chrishelgert: yes +1
also don't look to much into the implementation yet, its more about the concept at the current state :D

src/helpers/utils.ts
```diff
@@ -10,6 +10,7 @@ export function sendMessageToEditor(data: EditorMessage): void {
     ...data,
     from: 'live-preview',
     location: window.location.href,
+    version: '2.2.0', // TODO: set this automatically during publish (TOL-1178)
```
chrishelgert, line 13: 💥  We need that to prevent a breaking change on the webapp 💥

src/__tests__/resolveReference.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+        },
+      } as MessageEvent);
+
+      const result = await prom;
+
+      expect(result).toEqual(expected);
+    });
+
+    it('ignores different entries while waiting for the reference', async () => {
+      const prom = resolveReference({
```
aodhagan-cf, line 88: image

src/helpers/resolveReference.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  PromiseMap.set(referenceId, newPromise);
+
+  return newPromise;
+}
+
+export async function resolveReference(info: {
+  entityReferenceMap: EntityReferenceMap;
+  referenceId: string;
+  isAsset?: false;
```
aodhagan-cf, line 68: If you already have the function overload I wonder if there is much need for this boolean? 🤔

src/__tests__/resolveReference.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+import { resolveReference } from '../helpers/resolveReference';
+import { ASSET_TYPENAME } from '../types';
+
+vi.mock('../helpers/utils');
+
+describe('resolveReference', () => {
+  const listener = vi.spyOn(window, 'addEventListener');
+
+  beforeEach(() => {
+    vi.useFakeTimers();
```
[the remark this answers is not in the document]
chrishelgert: nope, but is best practice in jest to do it that way and as the vitest and jest are really similar I thought it would be also good here :D

src/helpers/resolveReference.ts
chrishelgert: yes you need it for the correct response type and to make the correct request to resolve the reference (.entries.get, vs .assets.get)

chrishelgert: ah get it now - is removed from the first one

src/__tests__/resolveReference.spec.ts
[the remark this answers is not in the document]
aodhagan-cf: Okay doke, looks good so. Just curious

src/graphql/entries.ts
```diff
@@ -55,15 +56,15 @@ export function updateEntry({
         locale,
       });
     } else if (field.type === 'Link') {
-      updateSingleRefField({
+      await updateSingleRefField({
```
aodhagan-cf, line 59: so now these are waiting for the promise to the editor to be resolved right?

src/rest/entities.ts
```diff
@@ -21,58 +21,54 @@ function getFieldName(contentType: ContentType, field: ContentType['fields'][num
...
-  const match = entityReferenceMap.get(updateFromEntryEditor.sys.id);
-  if (!match) {
-    sendMessageToEditor({
-      action: 'ENTITY_NOT_KNOWN',
-      referenceEntityId: updateFromEntryEditor.sys.id,
-      referenceContentType:
-        'linkType' in updateFromEntryEditor.sys ? updateFromEntryEditor.sys.linkType : undefined,
-    });
+): Promise<Reference | undefined | null> {
+  const { reference } = await resolveReference({
```
aodhagan-cf, line 30: Really like this new way, very clean

aodhagan-cf (approved):
Looks amazing! much cleaner solution with perf benefits 🤴👸

Do you think we will need to introduce an retry logic in future to the reference fetcher?

chrishelgert:
> Looks amazing! much cleaner solution with perf benefits 🤴👸
> 
> Do you think we will need to introduce an retry logic in future to the reference fetcher?

mh hopefully not, as there should be already a retry logic in the webapp part to resolve the reference 🤔 

what could be more possible that we need to implement a timeout, currently we would wait forever

PR #125, opened 2023-05-23
feat: remix example [TOL-1177]

YvesRijckaert:
Add a simple remix example with live preview

### DIFF
--- a/examples/remix/app/utils/preview-mode.server.ts
+++ b/examples/remix/app/utils/preview-mode.server.ts
@@ -1,0 +1,17 @@
+import { createCookie } from "@remix-run/node";
+
+import { parseCookie } from "./parse-cookie.server";
+
+export const previewModeCookie = createCookie("stage", {
+  path: "/",
+  sameSite: process.env.NODE_ENV !== "development" ? "none" : "lax",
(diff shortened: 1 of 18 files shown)


examples/remix/app/utils/preview-mode.server.ts
```diff
@@ -0,0 +1,17 @@
+import { createCookie } from "@remix-run/node";
+
+import { parseCookie } from "./parse-cookie.server";
+
+export const previewModeCookie = createCookie("stage", {
+  path: "/",
+  sameSite: process.env.NODE_ENV !== "development" ? "none" : "lax",
```
chrishelgert, line 7: 👍

examples/remix/jsconfig.json
aodhagan-cf, line 1: I'm very new to remix (this being the first remix code I've seen) but do you know why this is a jsconfig.json instead of a tsconfig?

YvesRijckaert: I think something got mixed up during the configuration, I changed it to a `tsconfig` now

chrishelgert (approved):

chrishelgert:
More general, what do you think about having a deployment for each example?
and added as content preview to one space

PR #130, opened 2023-05-26
feat: add version to package.json [TOL-1178]

2wce:


2wce:
yes @chrishelgert this needs to be merged first so that i can see if we now successfully update the package.json then we can complete the ticket in the next pr

chrishelgert (approved):

PR #138, opened 2023-05-30
feat(graphql): support for graphql aliases [TOL-1115]

chrishelgert:
Provide support for GraphQL aliases by parsing the provided query (DocumentNode)

### DIFF
--- a/src/graphql/entries.ts
+++ b/src/graphql/entries.ts
@@ -1,6 +1,13 @@
 import type { EntryProps } from 'contentful-management';
 
-import { isPrimitiveField, updatePrimitiveField, resolveReference, debug } from '../helpers';
+import {
+  isPrimitiveField,
+  updatePrimitiveField,
+  resolveReference,
+  clone,
+  debug,
+  generateTypeName,
+} from '../helpers';
 import {
   CollectionItem,
   SysProps,
@@ -11,9 +18,11 @@ import {
   UpdateReferenceFieldProps,
   UpdateEntryProps,
   isAsset,
+  GraphQLParams,
 } from '../types';
 import { updateAsset } from './assets';
-import { buildCollectionName, logUnrecognizedFields } from './utils';
+import { isRelevantField, updateAliasedInformation } from './queryUtils';
+import { buildCollectionName } from './utils';
 
 /**
  * Updates GraphQL response data based on CMA entry object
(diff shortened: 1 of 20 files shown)


src/graphql/entries.ts
```diff
@@ -30,21 +30,26 @@ export async function updateEntry({
...
+  const copyOfDataFromPreviewApp = clone(dataFromPreviewApp);
   const { fields } = contentType;
 
+
+  console.log(">> dataFromPreviewApp", dataFromPreviewApp)
+  console.log(">> aliases", aliases)
+
   logUnrecognizedFields(
-    fields.map((f) => f.apiName ?? f.name),
+    fields.map((field) => getAliasForFieldName(field.apiName ?? field.name)),
```
chrishelgert, line 47: I don't like that we have to do it twice, maybe we can include the log into the for loop

src/react.tsx
```diff
@@ -48,6 +49,23 @@ export function ContentfulLivePreviewProvider({
...
 
+interface Options {
+  locale?: string
+  /** GraphQL query related to the provided data */
+  query?: DocumentNode
+
+  skip?: boolean
+}
+
+export function useContentfulLiveUpdates<T extends Argument | null | undefined>(
```
chrishelgert, line 60: TODO: add to README

src/react.tsx
```diff
@@ -48,6 +49,23 @@ export function ContentfulLivePreviewProvider({
...
+  data: T,
+  options?: Options
+): T
+export function useContentfulLiveUpdates<T extends Argument | null | undefined>(
```
chrishelgert, line 64: should we deprecate this one?

src/graphql/entries.ts
```diff
@@ -131,8 +148,11 @@ async function updateReferenceEntryField(
 
   // TODO: kind of duplication with line 46, check if we can combine them
   for (const key in reference.fields) {
-    const value = reference.fields[key as keyof typeof reference.fields][locale];
+    if (!isRequestedField(key, typeName, merged, gqlParams)) {
```
chrishelgert, line 151: TODO: check if this works without a provided query (or if merged needs to be reference, but this breaks the function again)

src/graphql/entries.ts
```diff
@@ -30,51 +40,56 @@ export async function updateEntry({
...
-      });
-    } else if (field.type === 'Array' && field.items?.type === 'Link') {
-      await updateMultiRefField({
-        dataFromPreviewApp: copyOfDataFromPreviewApp,
-        updateFromEntryEditor,
-        name,
-        locale,
-        entityReferenceMap,
-      });
+    if (isRequestedField(name, typename, copyOfDataFromPreviewApp, gqlParams)) {
```
chrishelgert, line 54: With a query, we're only resolving requested information - should increase performance

src/react.tsx
aodhagan-cf: Probably

aodhagan-cf: ah right you did

src/graphql/entries.ts
```diff
@@ -30,52 +38,57 @@ export async function updateEntry({
...
```
aodhagan-cf, line 52: Any edge cases you can think of where they don't pass the query but want the data?

src/graphql/utils.ts
```diff
@@ -1,24 +1,132 @@
...
+  return params;
+}
+
+function getAliasForFieldName(name: string, typename: string, gqlParams?: GraphQLParams): string {
+  return gqlParams?.get(typename)?.alias.get(name) || name;
 }
 
-function removeCollection(name: string): string {
-  return name.replace(COLLECTION_SUFFIX, '');
+export function isRequestedField(
```
aodhagan-cf, line 98: Without the context I feel like this function could be hard to understand in future. Possibly something like
`isFieldInQuery` could be more explicit here

src/graphql/utils.ts
aodhagan-cf, line 1: Util to me feels like global utility helper functions, these new changes are specifically about modifying / understanding the graphql query provided. Maybe consider moving these functions to their own file, with a more specific name. `graphqlQueryUtils` even. Otherwise I think this file will grow a lot

aodhagan-cf: Maybe even a specific file for alias, and another for general query utility functions

src/graphql/utils.ts
```diff
@@ -1,24 +1,132 @@
...
+        );
+      }
     }
   }
+
+  return generated;
 }
 
-const COLLECTION_SUFFIX = 'Collection';
+export function parseGraphQLParams(query: DocumentNode): GraphQLParams {
```
aodhagan-cf, line 54: I'm a bit confused, could you please explain what this function does?

src/graphql/utils.ts
```diff
@@ -1,24 +1,132 @@
...
+
+interface GeneratedGraphQLStructure {
+  name: string;
+  alias?: string;
+  __typename: string;
+}
 
-export function logUnrecognizedFields(contentTypeFields: string[], data: Entity): void {
-  const recognized = new Set([...DEFAULT_CONTENT_TYPE_FIELDS, ...contentTypeFields]);
+function detectFieldAlias(
```
aodhagan-cf, line 29: `gatherAliasedFieldsFromSelections`?

src/graphql/entries.ts
chrishelgert: It should still be the same functionality for the behavior without gqlParams
- I only moved it from a check on every type level to one central place

src/graphql/utils.ts
chrishelgert: added documentation 👍

src/rest/entities.ts
```diff
@@ -142,32 +147,34 @@ export async function updateEntity(
...
+      await updateMultiRefField(
+        dataFromPreviewApp,
+        updateFromEntryEditor,
+        locale,
+        name as keyof Reference['fields'],
+        entityReferenceMap
+      );
     }
-  } catch (err) {
-    console.error('>> error', err);
```
chrishelgert, line 170: that was me once debugging and forgot to remove it 🙈

aodhagan-cf: hahaha

src/graphql/utils.ts
aodhagan-cf: Thank you for moving 🙏

aodhagan-cf (approved):
Looks good dude!

PR #139, opened 2023-05-30
fix: embedded entries rich text [TOL-1197]

YvesRijckaert:
Fix embedded entries in rich text with GraphQL

- [x] Merge web app PR to update the reference map (for newly created references): https://github.com/contentful/user_interface/pull/15675
- [ ] REST behaviour (covered in different PR)

### Demo videos
**Before**:
https://github.com/contentful/live-preview/assets/22968325/32a9a88c-eab8-4cc6-9e21-252217391a04
**After**:
https://github.com/contentful/live-preview/assets/22968325/7f370f5c-0a2e-436d-9367-f2ec87801f74

### DIFF
--- a/src/graphql/entries.ts
+++ b/src/graphql/entries.ts
@@ -101,24 +101,30 @@ async function processNode(
       const updatedReference = {
         sys: { id: id, type: 'Link', linkType: node.data.target.sys.type },
       };
-      // Use the updateReferenceEntryField function to resolve the entity reference
-      const ref = await updateReferenceEntryField(
-        null,
-        updatedReference,
-        entityReferenceMap,
-        locale
-      );
+      let ref;
+
+      // Use the updateReferenceEntryField or updateReferenceAssetField function to resolve the entity reference
+      if (node.data.target.sys.linkType === 'Entry') {
+        ref = await updateReferenceEntryField(null, updatedReference, entityReferenceMap, locale);
+      } else if (node.data.target.sys.linkType === 'Asset') {
+        ref = await updateReferenceAssetField({
+          referenceFromPreviewApp: null,
+          updatedReference,
+          entityReferenceMap,
+          locale,
+        });
+      }
 
       // Depending on the node type, assign the resolved reference to the appropriate array
       switch (node.nodeType) {
         case 'embedded-entry-block':
-          entries.block.push(ref);
+          if (ref) entries.block.push(ref);
           break;
         case 'embedded-entry-inline':
-          entries.inline.push(ref);
+          if (ref) entries.inline.push(ref);
           break;
         case 'embedded-asset-block':
-          assets.block.push(ref);
+          if (ref) assets.block.push(ref);
           break;
       }
     }
--- a/src/graphql/entries.ts
+++ b/src/graphql/entries.ts
@@ -214,6 +214,16 @@ async function updateReferenceEntryField(
       if (value.nodeType === 'document') {
         // richtext
         merged[key] = { json: value };
+        // Process the Rich Text field
+        const entries: RichTextLink = { block: [], inline: [] };
+        const assets: RichTextLink = { block: [], inline: [] };
+        for (const node of value.content) {
+          await processNode(node, entries, assets, entityReferenceMap, locale);
+        }
+        merged[key].links = {
+          entries: isEntityLinkEmpty(entries) ? undefined : entries,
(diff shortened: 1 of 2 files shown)


src/graphql/entries.ts
```diff
@@ -77,18 +78,82 @@ export async function updateEntry({
...
+      };
+
+      // Iterate over each node in the rich text content
+      for (const node of dataFromPreviewApp[name].json.content) {
+        // Check if the node is an embedded entry or asset, block or inline
+        if (
+          node.nodeType === 'embedded-entry-block' ||
+          node.nodeType === 'embedded-entry-inline' ||
+          node.nodeType === 'embedded-asset-block' ||
+          node.nodeType === 'embedded-asset-inline'
```
MayaGillilanContentful, line 116: We can probably just simplify this to `node.nodeType.includes('embedded')`

src/graphql/entries.ts
```diff
@@ -77,18 +78,82 @@ export async function updateEntry({
...
+        ) {
+          if (node.data && node.data.target && node.data.target.sys) {
+            const id = node.data.target?.sys.id || '';
+            const updatedReference = {
+              sys: { id: id, type: 'Link', linkType: node.data.target.sys.type },
+            };
+
+            // Use the updateReferenceEntryField function to resolve the entity reference
+            const ref = await updateReferenceEntryField(
```
MayaGillilanContentful, line 125: Why this instead of resolveReference?

src/graphql/entries.ts
```diff
@@ -77,18 +78,82 @@ export async function updateEntry({
...
+              case 'embedded-asset-inline':
+                assets.inline.push(ref);
+                break;
+            }
+          }
+        }
+      }
+
+      // Finally, assign the arrays of resolved references to the 'links' property of the field in the preview app data
+      dataFromPreviewApp[name].links = {
```
MayaGillilanContentful, line 152: Should we strip empty arrays out, or just not care?

src/graphql/entries.ts
```diff
@@ -77,18 +78,76 @@ export async function updateEntry({
...
+interface RichTextLink {
+  block: (Entity & CollectionItem)[];
+  inline: (Entity & CollectionItem)[];
+}
+
+function isEntityLinkEmpty(obj: RichTextLink) {
+  return Object.values(obj).every((arr) => arr.length === 0);
+}
+
+async function updateRichTextField({
```
chrishelgert, line 90: can we also call it there now?
https://github.com/contentful/live-preview/pull/139/files#diff-0b6094972cd316ac3e76d7602561d8d08032a3e09d809cbf03e8574cc91b3062R198
Otherwise, it will not work for nested references

src/graphql/entries.ts
```diff
@@ -77,18 +78,76 @@ export async function updateEntry({
...
+          };
+          // Use the updateReferenceEntryField function to resolve the entity reference
+          const ref = await updateReferenceEntryField(
+            null,
+            updatedReference,
+            entityReferenceMap,
+            locale
+          );
+          // Depending on the node type, assign the resolved reference to the appropriate array
+          switch (node.nodeType) {
```
chrishelgert, line 128: Does this mean we will need to implement also resource links here? 🤔

YvesRijckaert: I don't think we support resource links as embedded entries in rich text yet right?
But once we do I guess we should test it (I think resource links open in a new tab so not sure if we need to do anything)

YvesRijckaert: I'll create a follow-up ticket for this then, I think we should try to merge this one in sooner rather than later since it's a few users have brought it up already

src/graphql/entries.ts
YvesRijckaert: Done 👍

src/graphql/entries.ts
```diff
@@ -77,18 +78,105 @@ export async function updateEntry({
...
+          updatedReference,
+          entityReferenceMap,
+          locale,
+        });
+      }
+
+      // Depending on the node type, assign the resolved reference to the appropriate array
+      switch (node.nodeType) {
+        case 'embedded-entry-block':
+          if (ref) entries.block.push(ref);
```
chrishelgert, line 121: we could check if there is a ref before going into the switch, and just push inside the switch/case

chrishelgert (approved):
Two tiny comments, otherwise look good 👍 
When implementing REST, we should check if we can share the code and add unit tests

PR #141, opened 2023-06-02
fix: embedded entries rich text REST [TOL-1197]

YvesRijckaert:
Fix embedded entries in rich text with REST


chrishelgert (approved):
Tested and works out well 👍

PR #151, opened 2023-06-07
docs(draft-mode): update workaround to share the draft mode with an iframe

chrishelgert:
- Add a note for the workaround in the README.md
- Adjust the workaround to only update the necessary cookie


aodhagan-cf (approved):
LGTM

PR #172, opened 2023-06-22
fix(graphql): fields not updating when creating a new entry [TOL-1251]

chrishelgert:
- Fields were not updated for GraphQL if the entry was newly generated.
This was because of the `isRelevantField` implementation, now it updates every field, which produces some adjustments in the tests.
- "cannot read content of null" due a missing `null` type on `richTextNode`

### DIFF
--- a/src/graphql/__tests__/queryUtils.test.ts
+++ b/src/graphql/__tests__/queryUtils.test.ts
@@ -74,35 +74,20 @@ describe('updateAliasedInformation', () => {
 });
 
 describe('isRelevantField', () => {
-  it('validates the information based on the provided data', () => {
-    const data = {
-      sys: { id: '1' },
-      name: 'Test',
-      content: {},
-      topSectionCollection: { items: [] },
-    };
-
-    // direct match => true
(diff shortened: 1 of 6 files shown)


src/graphql/__tests__/queryUtils.test.ts
```diff
@@ -74,35 +74,20 @@ describe('updateAliasedInformation', () => {
...
-
-    // direct match => true
-    expect(isRelevantField('content', 'pageCollection', data)).toBeTruthy();
-    // collection match => true
-    expect(isRelevantField('topSection', 'pageCollection', data)).toBeTruthy();
-    // unknown alias => false
-    expect(isRelevantField('internalName', 'pageCollection', data)).toBeFalsy();
-    // unknown field => false
-    expect(isRelevantField('header', 'pageCollection', data)).toBeFalsy();
+  it('returns true because no are queryParams provided', () => {
```
MayaGillilanContentful, line 77:
```suggestion
  it('returns true because no queryParams are provided', () => {
```

MayaGillilanContentful (approved):

PR #186, opened 2023-06-28
fix: hyperlink references in REST and update tests [TOL-1290]

YvesRijckaert:
- fixes an issue with entry-hyperlink and asset-hyperlink not resolving correctly
- the previous PR (https://github.com/contentful/live-preview/pull/181) introduced tests, but these were false positives.. 😬 I refactored the tests and now they should capture all the possible embedded entries (including entry-hyperlink and asset-hyperlink)

### DIFF
--- a/src/rest/__tests__/entities.test.ts
+++ b/src/rest/__tests__/entities.test.ts
@@ -4,35 +4,14 @@ import { describe, it, expect, vi, afterEach, beforeEach, Mock } from 'vitest';
 import contentTypeAsset from '../../__tests__/fixtures/contentTypeAsset.json';
 import { clone, resolveReference } from '../../helpers';
 import { EntityReferenceMap } from '../../types';
-import { updateEntity, updateRef } from '../entities';
+import { updateEntity } from '../entities';
 import contentTypeEntry from './fixtures/contentType.json';
 import dataFromPreviewApp from './fixtures/dataFromPreviewApp.json';
 import asset from './fixtures/updateAssetFromEntryEditor.json';
 import entry from './fixtures/updateFromEntryEditor.json';
 
 const EN = 'en-US';
-const referenceId = 'referenceId';
-const richTextFieldName = 'richTextFieldName';
-const richTextContent = {
-  nodeType: 'document',
-  content: [
-    {
-      nodeType: 'embedded-entry-block',
-      data: {
-        target: {
-          sys: {
-            id: 'foo',
-            type: 'Entry',
-          },
-          fields: {
-            headline: { [EN]: 'Headline' },
-          },
-        },
-      },
-    },
-  ],
-  data: {},
-};
+const referenceWithRichTextId = 'referenceWithRichTextId';
 
 function patchField<T extends EntryProps | AssetProps>(
   originalData: T,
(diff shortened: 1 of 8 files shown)


src/rest/entities.ts
```diff
@@ -134,7 +134,11 @@ async function resolveRichTextLinks(
   entityReferenceMap: EntityReferenceMap,
   locale: string
 ) {
-  if (node.nodeType.includes('embedded')) {
+  if (
+    node.nodeType.includes('embedded') ||
+    node.nodeType === 'entry-hyperlink' ||
```
YvesRijckaert, line 139: Will investigate if we have to add this in GraphQL as well, will create a followup PR if necessary

src/rest/__tests__/entities.test.ts
```diff
@@ -254,31 +390,14 @@ describe('Update REST entry', () => {
...
-      const result = (await updateRef(
-        dataFromPreviewApp,
-        updateFromEntryEditor,
-        EN,
-        defaultEntityReferenceMap
-      )) as unknown as EntryProps;
-
-      expect(result).not.toBeNull();
-      expect(result.fields[richTextFieldName]).toEqual(richTextContent);
+    it('resolves reference with rich text with embedded references', async () => {
```
chrishelgert, line 393:
```suggestion
    it('resolves reference for rich text with embedded references', async () => {
```

src/rest/__tests__/entities.test.ts
```diff
@@ -66,16 +45,110 @@ describe('Update REST entry', () => {
     },
   } as unknown as AssetProps;
 
+  const unresolvedRichTextLinks = {
```
chrishelgert, line 48: as the test file is going to be a little bit long, should we extract the test data into a fixture?

src/rest/entities.ts
z0al: What about asset and resource links? Also, at this point maybe it's better to use `@contentful/rich-text-types` to get node type ids

YvesRijckaert: assets are supported (`node.nodeType.includes('embedded')` includes the embedded-asset-block + we have the asset-hyperlink)
I'll refactor it to use the types from the package. Not sure about resource links, since they will open in a new tab, so live updates will not work there?

z0al: > assets are supported (node.nodeType.includes('embedded') includes the embedded-asset-block + we have the asset-hyperlink)

I'm sorry. I was probably too busy to read the PR carefully. You're right.

z0al: > Not sure about resource links, since they will open in a new tab, so live updates will not work there

If resources are not relevant, do we need to maybe exclude `embedded-resource-block` from being included with `includes('embedded')`?

YvesRijckaert: I am not sure, do you mind if we create a follow up ticket for that? We need to investigate either way how embedded resources links work in live preview, (we haven't tested it out yet)

YvesRijckaert: https://contentful.atlassian.net/browse/TOL-1293

chrishelgert (approved):
Looks fine, what's the impact on the bundle size with the new package included?

Issue #179, opened 2023-06-26
Rich text fields with hyperlink don't appear to work


nrandell:
If you've got a rich text field with a hyperlink, it is reported as 
```
links {
  entries {
    hyperlink {
      ... data for the hyperlink
    }
  }
}
```

But the live preview doesn't appear to support hyperlink, so when an update occurs, the hyperlink field is removed. I think it's all around this code https://github.com/contentful/live-preview/blob/e11dfe2627fb0eb9a1c6f91f4b4fc7d14eb6d04d/src/graphql/entries.ts#L174 and around `processRichTextField`

chrishelgert:
Thank you for reporting the issue, we will have a look

PR #187, opened 2023-06-30
fix: graphql embedded entries hyperlinks [TOL-1262]

YvesRijckaert:
- fix an issue with entry and asset hyperlinks with graphql (fixes https://github.com/contentful/live-preview/issues/179)
- use rich text types package to determine the types
- add missing debug warn in REST version when we are handling an unknown embedded entry type

https://github.com/contentful/live-preview/assets/22968325/b792f2b2-64e5-4c6a-bc87-e6e4894c605a

### DIFF
--- a/src/graphql/entries.ts
+++ b/src/graphql/entries.ts
@@ -1,3 +1,4 @@
+import { BLOCKS, INLINES } from '@contentful/rich-text-types';
 import type { EntryProps } from 'contentful-management';
 
 import {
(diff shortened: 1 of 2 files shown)


aodhagan-cf (approved):
Really nice! A test would also be great for this if you get the chance 🎉

YvesRijckaert:
> Really nice! A test would also be great for this if you get the chance 🎉

I always forget! Will have a look 🐥

YvesRijckaert:
@aodhagan-cf I'll merge this one to unblock the customers
I'll create a follow up ticket for adding tests (won't be able to focus on that soon I think)

PR #195, opened 2023-07-04
fix: circular references max depth REST [TOL-1260]

YvesRijckaert:
Fixing circular references by adding max depth in REST API


chrishelgert (approved):
LGTM

PR #196, opened 2023-07-04
chore: adjust type error

chrishelgert:
Don't know why it was not complaining on the previous PR, but I get a error locally about a missing property - let's adjust this.


z0al (approved):
LGTM. Though a "fix" would be more appropriate

chrishelgert:
> LGTM. Though a "fix" would be more appropriate

As the previous one was also a chore - it was not published yet so triggering a release for this one seems to be odd
What do you think?

PR #198, opened 2023-07-05
chore: migrate to monorepo

chrishelgert:
Migrate the SDK to mono repo so we can easily share features and have more flexibility on the solutioning


package.json
```diff
@@ -1,97 +1,48 @@
...
-    "test:ci": "vitest run",
-    "coverage": "vitest run --coverage",
+    "lint": "lerna run lint",
+    "tsc": "lerna run tsc",
+    "test": "lerna run --stream test",
+    "test:ci": "lerna run test:ci",
+    "coverage": "lerna run coverage",
     "prepare": "husky install",
-    "presemantic-release": "yarn build",
+    "build": "lerna run build",
```
2wce, line 25: we could add turborepo like we have in field editors & use it for the build command it makes it quite fast once we start to add more packages

2wce (approved):

PR #208, opened 2023-07-10
fix: circular references and infinite loop problems

YvesRijckaert:
### This PR fixes multiple issues:
- We were using the`useDeepCompareEffectNoCheck` package, which has a dependency on dequal package. Dequal has an issue with infinite loops in circular references: https://github.com/lukeed/dequal/issues/19 . I removed this package in favour of the is isEqual package from lodash. This one seems to work better with circular references.
- Even with the previously introduced depth limitation for REST (https://github.com/contentful/live-preview/pull/195), you could still encounter infinite loops if an entity links to itself directly or indirectly. Now, when a circular reference is detected, it does not update the reference further. Instead, it returns the the already updated reference from the `entityReferenceMap`
- There was an issue with resource links being wrongly detected as normal references. I've added some todo's for follow up PRs to implement the live update behaviour for resource links.

### Screenshots
Before:
Screenshot 2023-07-13 at 10 53 15

After:
Screenshot 2023-07-13 at 10 56 54

### DIFF
--- a/packages/live-preview-sdk/package.json
+++ b/packages/live-preview-sdk/package.json
@@ -51,10 +51,11 @@
   "dependencies": {
     "@contentful/rich-text-types": "^16.2.0",
     "@contentful/visual-sdk": "^1.0.0-alpha.1",
-    "use-deep-compare-effect": "^1.8.1"
+    "lodash.isequal": "^4.5.0"
   },
   "devDependencies": {
     "@testing-library/react": "14.0.0",
+    "@types/lodash.isequal": "^4.5.6",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
   },
   "homepage": "https://github.com/contentful/live-preview#readme",
   "dependencies": {
+    "lodash": "^4.17.21",
```
chrishelgert, line 55: should we maybe only import `lodash.isEqual`? instead of having the whole lodash a dependency

chrishelgert (approved):
LGTM

PR #214, opened 2023-07-10
feat(sdk): extract entity store for re-usability

chrishelgert:
- Extract the EntityStore for loading entities from the editor into separate modules
- Use it for resolving references
- Adjust the first steps for later message optimizations (from -> source, action -> method) for unified messaging

### DIFF
--- a/packages/visual-sdk/src/store/EditorEntityStore.ts
+++ b/packages/visual-sdk/src/store/EditorEntityStore.ts
@@ -1,41 +1,58 @@
 import type { Asset, Entry } from 'contentful';
+
 import { EntityStore } from './EntityStore';
 
-type SendMessage = (params: EditorEntityStoreMessage) => void;
+export type RequestEntitiesMessage = {
+  entityIds: string[];
+  entityType: 'Asset' | 'Entry';
+  locale: string;
+};
 
-export type EditorEntityStoreMessage = {
-  action: 'REQUEST_ENTITIES';
-  entityIds: string[],
-  entityType: 'Asset' | 'Entry',
-  locale: string,
-}
-
-export type EditorEntityStoreReceivedMessage = {
-  action: 'REQUESTED_ENTITIES';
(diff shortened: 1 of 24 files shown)


packages/visual-sdk/src/store/EditorEntityStore.ts
```diff
@@ -0,0 +1,169 @@
...
+      const entry = this.entitiesMap.get(id);
+      if (!entry) {
+        missing.push(id);
+      }
+    }
+
+    return missing;
+  }
+
+  private async fetchEntity(ids: string[]): Promise<Array<Entry>>;
```
YvesRijckaert, line 83: I find this code a bit confusing, could explain what's happening here? Why do we have 3 times `fetchEntity`? Can we not have one generic function and then fetchAsset?

packages/visual-sdk/src/store/EntityStore.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    if (missingEntityIds.length) {
+      throw new Error(`Missing entity in the store (${missingEntityIds.join(',')})`);
+    }
+
+    return entity as Array<Asset | Entry>;
+  }
+
+  public async fetchAsset(id: string): Promise<Asset | undefined> {
```
YvesRijckaert, line 61: should these methods be async? Aren't we just getting something from a map?

YvesRijckaert (approved):
Just two questions, looks good!

packages/visual-sdk/src/store/EditorEntityStore.ts
chrishelgert: Ah that's for typing ([function overload](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads))- so you get the correct type

if you call it without  `isAsset` you get the `Entry` type
if you call it with `isAsset` you get the `Asset` type
and the third one is then the implementation, which needs to implement both overloads

PR #218, opened 2023-07-13
chore: allow react 17 for peerDependencies

chrishelgert:
As we don't use any react 18 specific code, there speaks nothing against adding version 17 of react to the peer dependencies


YvesRijckaert (approved):
Should we add to the docs that you need react 17 or higher?

YvesRijckaert (changes requested):
I forgot to run yarn after rebasing in my PR, could you do that as part of this PR? 😅

PR #231, opened 2023-07-19
fix(types): optimize export of typings

chrishelgert:
Extract the enum for the Post Messages and the correlated message types from './types' as a `enum` is technically more than a type and to be able to easily share them
Optimizes also the exports so you don't need to import from 'dist/types' anymore

### DIFF
--- a/packages/live-preview-sdk/src/__tests__/validateMessages.spec.ts
+++ b/packages/live-preview-sdk/src/__tests__/validateMessages.spec.ts
@@ -1,0 +1,34 @@
+import { describe, it, expect } from 'vitest';
+
+import { LIVE_PREVIEW_EDITOR_SOURCE } from '../constants';
+import { isValidMessage } from '../helpers/validateMessage';
+
+describe('isValidMessage', () => {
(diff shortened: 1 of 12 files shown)


packages/live-preview-sdk/src/__tests__/validateMessages.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+  it('should return false because neither event.from and event.source is provided', () => {
+    expect(isValidMessage({ data: { any: 'data' } } as MessageEvent)).toBeFalsy();
+  });
+  it('should return true because the event.source is valid', () => {
+    expect(
+      isValidMessage({ data: { any: 'data', source: LIVE_PREVIEW_EDITOR_SOURCE } } as MessageEvent)
+    ).toBeTruthy();
+  });
+  it('should return false because the event.from is valid', () => {
```
YvesRijckaert, line 29:
```suggestion
  it('should return true because the event.from is valid', () => {
```

YvesRijckaert (approved):

PR #233, opened 2023-07-21
feat: support fallback locales

chrishelgert:
The editor will now resolve the locale and also handles the fallback logic of locales.
This adopts the SDK to work with the new structure

### DIFF
--- a/packages/live-preview-sdk/package.json
+++ b/packages/live-preview-sdk/package.json
@@ -67,6 +67,7 @@
     "jsdom": "^22.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "type-fest": "^4.0.0",
     "typescript": "^5.0.2",
     "vite": "^4.1.0",
     "vite-plugin-dts": "2.3.0",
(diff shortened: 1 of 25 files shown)


packages/live-preview-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/live-preview",
-  "version": "2.5.4",
+  "version": "2.6.0-alpha.0",
```
chrishelgert, line 3: Revert this before merging

packages/visual-sdk/package.json
```diff
@@ -44,7 +44,7 @@
     "tsc": "tsc --noEmit",
     "test": "jest --watch",
     "test:ci": "jest",
-    "coverage": "vitest --coverage",
+    "coverage": "jest --coverage",
```
YvesRijckaert, line 47: why are we using jest again instead of vitest? :)

YvesRijckaert (approved):
LGTM

packages/visual-sdk/package.json
chrishelgert: the visual sdk uses swc for the build and not vite, and therefore it's easier to setup jest with swc :) 
otherwise we would need to have still the vite dependency, add having a vite config

PR #255, opened 2023-08-02
fix(live-preview): provide better information for status modal

chrishelgert:
- Provide more useful information for the status modal inside the live preview editor


packages/live-preview-sdk/src/messages.ts
```diff
@@ -44,8 +45,10 @@ export type ConnectedMessage = {
   /** @deprecated use method instead */
   action: LivePreviewPostMessageMethods.IFRAME_CONNECTED | LivePreviewPostMessageMethods.CONNECTED;
   connected: true;
-  tags: number;
   locale: string;
+  /** @deprecated use taggedElementCount instead */
+  tags: number;
```
chrishelgert, line 50: `tags` is not really meaningful, so I would renamed it to `taggedElementCount`

2wce (approved):

PR #264, opened 2023-08-08
fix(live-preview): adjust missing data for the status modal

chrishelgert:
`this.inspectorMode?.getTaggedElements()` as the inspectorMode could be undefined, we don't always send the tagged elements and therefore the analytics for the statusmodal can be faulty.

### DIFF
--- a/packages/live-preview-sdk/src/index.ts
+++ b/packages/live-preview-sdk/src/index.ts
@@ -126,17 +126,20 @@ export class ContentfulLivePreview {
       pollUrlChanges(() => {
         sendMessageToEditor(LivePreviewPostMessageMethods.URL_CHANGED, {
           action: LivePreviewPostMessageMethods.URL_CHANGED,
-          taggedElementCount: this.inspectorMode?.getTaggedElements().length,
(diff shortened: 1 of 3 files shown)


2wce (approved):

packages/live-preview-sdk/src/index.ts
```diff
@@ -126,17 +126,19 @@ export class ContentfulLivePreview {
...
       });
 
       // tell the editor that there's a SDK
       sendMessageToEditor(LivePreviewPostMessageMethods.CONNECTED, {
         action: LivePreviewPostMessageMethods.CONNECTED,
         connected: true,
-        tags: this.inspectorMode?.getTaggedElements().length,
-        taggedElementCount: this.inspectorMode?.getTaggedElements().length,
+        tags: document.querySelectorAll(`[${TagAttributes.ENTRY_ID}]`).length,
+        taggedElementCount: document.querySelectorAll(`[${TagAttributes.ENTRY_ID}]`).length,
```
z0al, line 138: Nit: both queries are identical. Could you just call it once before the call?

z0al (approved):

PR #280, opened 2023-08-16
feat: expose opening entry [TOL-1237]

YvesRijckaert:
Exposing and decoupling the logic for opening an entry in the Contentful UI from the specific inspector mode class

### DIFF
--- a/packages/live-preview-sdk/src/messages.ts
+++ b/packages/live-preview-sdk/src/messages.ts
@@ -4,6 +4,7 @@ import type { Asset, Entry } from 'contentful';
 import type { SysLink } from 'contentful-management';
 
 import type { LIVE_PREVIEW_EDITOR_SOURCE, LIVE_PREVIEW_SDK_SOURCE } from './constants';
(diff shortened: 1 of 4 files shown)


aodhagan-cf (approved):
LGTM 🎉

packages/live-preview-sdk/src/messages.ts
```diff
@@ -159,3 +160,12 @@ export type MessageFromEditor = (
   from: 'live-preview';
   source: typeof LIVE_PREVIEW_EDITOR_SOURCE;
 };
+
+export function openEntryInEditorUtility(fieldId: string, entryId: string, locale: string): void {
```
z0al, line 164: You probably need to add envId & spaceId here

YvesRijckaert: I tried it out without and it works fine🤔 I guess it just wouldn't work with x-space references, but that counts for everything in the SDK.

z0al: It can also wait btw

Issue #217, opened 2023-07-13
Add a vanilla JS example


groenroos:
Could an example project please be provided in the `./examples` folder for a complete vanilla JS implementation? Each implementation in there currently seems to use React.

Alternatively (or ideally additionally) the documentation for the vanilla JS approach could be expanded, as the code examples provided include multiple undefined variables (`id`, `title`, `data`, etc) with no explanation what they are, where they should come from, or what they should contain.

Related to #152

cachrisman:
I agree that the [Vanilla JS documentation](https://github.com/contentful/live-preview#vanilla-javascript) in the README is lacking, but here's what I've been able to figure out:

Before you initialize the live preview you'll want to make whatever call (Content Preview API or Content GraphQL API) to get the entries and assets that make up the page being requested. Then for each entry+field on which you want to use the inspector mode to show an edit button in the preview you'll need to run `ContentfulLivePreview.getProps({ entryId: id, fieldId: title });` and then also `ContentfulLivePreview.subscribe({data, locale, callback});` to get updates when a field is edited in Contentful. 

- `id` is the entry ID
- `data` is the JSON response from the CPA for an entry/asset or an array of entries (or assets)
- `callback` is a function to be called when the entry/asset is updated in Contentful to tell the frontend/live preview to update the rendering. The callback is what makes the frontend update almost instantaneously when typing in a field in the editor.

I hope this is helpful!!

dreysolano:
Another issue to note is that older projects that use Webpack 4 will encounter build failures for both latest `contentful` (v10.4.1) and `@contentful/live-preview` packages, both requires adding `babel-loader` to your Webpack config. The below is for a Nuxt2 project, but should provide the basis for what you'll need.

```
config.module.rules.push({
  test: /\.[cm]?js$/,
  exclude: /node_modules\/(?!(contentful\/dist|@contentful\/live-preview)\/).*/,
  use: {
    loader: 'babel-loader',
    options: {
      presets: [
        [
          '@babel/preset-env',
          {
            targets: {
              browsers: ['last 2 versions', 'ie >= 11'],
            },
          },
        ],
      ],
      plugins: ['@babel/plugin-proposal-class-properties'],
    },
  },
})
```

The @babel/plugin-proposal-class-properties dependency is required for a `contentful@10.4.1` issue around static class properties. Note that I also had to use a negative lookahead assertion in the exclude regex to include the necessary package files for transpiling. Hope this helps.

Issue #152, opened 2023-06-07
VanillaJS live update instructions appear to be incomplete


jacksonopp:
The instructions for the the live update instructions do not mention what the values for `callback` and `data` are, or what they represent. Looking into the type definitions do not help either, as `data` essentially resolves to a `Record<any, any>` and `callback` is essentially `(data: Record<any, any>) => void`

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Live Preview Example</title>
    <link
      rel="stylesheet"
      type="text/css"
      href="./node_modules/@contentful/live-preview/dist/style.css"
    />
    <script type="module">
      import { ContentfulLivePreview } from './node_modules/@contentful/live-preview/dist/index.js';

      const locale = 'en-US';

      ContentfulLivePreview.init({ locale });

      /**
       * Subscribe to data changes from the Editor, returns a function to unsubscribe
       * Will be called once initially for the restored data
       */
      const unsubscribe = ContentfulLivePreview.subscribe({
        data, // <- This is undefined within this script
        locale,
        callback, // <- This is undefined within this script
      });
    </script>
  </head>
  <body></body>
</html>
```

YvesRijckaert:
I will reopen this ticket, we will add a vanilla JS example soon and update the docs.

YvesRijckaert:
Please have a look at the updated documentation for vanilla js and our new example in the examples directory :)

PR #287, opened 2023-08-23
docs: vanilla js example [TOL-1416]

YvesRijckaert:
- add a vanilla js example project with the SDK installed
- update the docs

Fixes:
- https://github.com/contentful/live-preview/issues/217
- https://github.com/contentful/live-preview/issues/152

### DIFF
--- a/examples/vanilla-js/app.js
+++ b/examples/vanilla-js/app.js
@@ -1,0 +1,75 @@
+import { ContentfulLivePreview } from '@contentful/live-preview';
+import { createClient } from 'contentful';
+import dotenv from 'dotenv';
+
+// Configuration object
+const CONFIG = {
+  locale: 'en-US',
+  entryId: 'W9z8t0uXz3CMFjp4sw5zj',
+  fields: ['title'],
+  subscriptions: [],
+  debugMode: true,
+};
+
+document.addEventListener('DOMContentLoaded', initialize);
+
+function initialize() {
+  dotenv.config();
+
+  const client = createClient({
+    space: process.env.CONTENTFUL_SPACE_ID,
+    accessToken: process.env.CONTENTFUL_ACCESS_TOKEN,
+    host: 'preview.contentful.com',
+  });
+
+  ContentfulLivePreview.init({ locale: CONFIG.locale });
+
+  client
+    .getEntry(CONFIG.entryId)
+    .then((entry) => {
+      CONFIG.fields.forEach((fieldId) => {
+        displayFieldData(entry, client, fieldId);
+        setupLivePreview(entry, fieldId);
+      });
+    })
+    .catch((err) => console.error(err));
+}
+
+function findElementByDataAttributes(entryId, fieldId) {
+  return document.querySelector(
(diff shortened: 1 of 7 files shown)


examples/vanilla-js/app.js
```diff
@@ -0,0 +1,80 @@
...
+  setAttributes(domElement, props);
+
+  domElement.textContent = entry.fields[fieldId];
+}
+
+function setupLivePreview(entry, fieldId) {
+  const callback = (updatedData) => {
+    const domElement = document.getElementById(fieldId);
+    if (domElement && updatedData.fields && updatedData.fields[fieldId]) {
+      domElement.textContent = updatedData.fields[fieldId];
```
chrishelgert, line 69: should we check if it's text? and otherwise call a function to wait 5s and reload?
Guess that could be an use-case that would work for some

chrishelgert: ah see, our example is only for one text 🤔

examples/vanilla-js/index.html
```diff
@@ -0,0 +1,15 @@
...
+    <title>Contentful Live Preview Demo</title>
+    <link
+      rel="stylesheet"
+      type="text/css"
+      href="./node_modules/@contentful/live-preview/dist/style.css"
+    />
+    <script src="./app.js"></script>
+  </head>
+  <body>
+    <h1 id="title"></h1>
```
chrishelgert, line 13: should we add the inspectorMode attributes? and use them to find the element?

examples/vanilla-js/app.js
```diff
@@ -0,0 +1,80 @@
...
+    if (key.startsWith('data-')) {
+      const formattedName = key.split('data-')[1].replace(/-([a-z])/g, (m, w) => w.toUpperCase());
+      element.dataset[formattedName] = value;
+    } else {
+      element.setAttribute(key, value);
+    }
+  }
+}
+
+function displayFieldData(entry, client, fieldId) {
```
chrishelgert, line 52: Not sure about this one, or if we should then create the attributes manually in the HTML file, see https://github.com/contentful/live-preview/pull/287/files#r1303901947

YvesRijckaert: We had it like that in the documentation before, but I am also happy to go with the manual approach. I'll change it in both places then

chrishelgert: I think in the documentation it's fine if we mention the function - for example, if you have a node server and render the HTML on the server, you could use the helper to create the tags

examples/vanilla-js/app.js
YvesRijckaert: Yes all our examples are just for text fields. To keep things simple.
I can add a check if it's text though

examples/vanilla-js/app.js
YvesRijckaert: I'v changed it to set the attributes manually in the HTML. Does it look good now?

chrishelgert (approved):

PR #288, opened 2023-08-25
fix(graphql): providing the query no longer prevents updates [TOL-1409]

chrishelgert:
The typename for fields inside a collection was not correctly generated, and therefore they were not identified correctly and in `isRelevantField` they were ignored.
We now generate the typename correctly instead of  `items`, which enables updates again.

Migrated the nextjs-graphql example to use graphql-request and GraphQL Documents, which allows us to also demo the usage with a provided query


YvesRijckaert (approved):
LGTM

PR #301, opened 2023-08-29
feat(live-preview): add functionality to subscribe to the save event of an entity

chrishelgert:
- [x] Expose the save event for subscription
- [x] Add an example with the save event (SSR) only
- [x] Add it to the README


examples/next-13-app-router-ssr/app/page.tsx
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export default async function Home() {
+  const { isEnabled } = draftMode();
+
+  const posts = await getAllPostsForHome(isEnabled);
+
+  if (!posts || posts.length === 0) {
+    return (
+      <>
```
aodhagan-cf, line 17: nit: why the fragment here?

packages/live-preview-sdk/src/index.ts
```diff
@@ -201,6 +226,13 @@ export class ContentfulLivePreview {
 
     openEntryInEditorUtility(fieldId, entryId, locale || this.locale);
   }
+
+  /**
+   * Returns a list of tagged entries on the page
+   */
+  static getEntryList(): string[] {
+    return getEntryList();
```
aodhagan-cf, line 234: It might be worth renaming one of these, it looks a bit weird as is

packages/live-preview-sdk/src/utils.ts
aodhagan-cf, line 1: As a personal preference I would rename this from `util` to some more related to the functions.

`field-tag-utils`

and then the function in here maybe something like `getAllTaggedEntries`?

examples/next-13-app-router-ssr/public/_live-preview.ts
aodhagan-cf, line 1: **general question**: Is it possible to miss any updates with this?

For example:
1. user types "Hello"
2. auto save occurs
3. window reloading
4. user types "World"
5. page reloads

can we possibly lose the "World"

aodhagan-cf: haha no worries, just curious

examples/next-13-app-router-ssr/app/layout.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Inter } from 'next/font/google';
+import { draftMode } from 'next/headers';
+import Script from 'next/script';
+
+import '@contentful/live-preview/style.css';
+
+const inter = Inter({ subsets: ['latin'] });
+
+export const metadata: Metadata = {
+  title: 'Create Next App',
```
aodhagan-cf, line 11: Maybe something more relevant to the example here?

aodhagan-cf (approved):
This looks great!

PR #306, opened 2023-09-05
fix(example): next 13 with app router

Algram:
Hi there,

I am currently setting up the live preview mode for Contentful. The `live-updates` are working, but the inspector mode was missing the css import in the next 13 example app:
https://github.com/contentful/live-preview/tree/main/examples/next-13-app-router-graphql

I have also deleted the `globals.css` and the `prepare` step in the package.json, since they where both unused and preventing the app from starting.

:v:


chrishelgert (approved):

chrishelgert:
Thanks for your contribution 🥇

PR #377, opened 2023-10-23
fix: fixing circular ref updates by using store of already fetched data []

aodhagan-cf:
For circular references we were getting links instead of resolved entities when the same references is linked twice in the page

### DIFF
--- a/packages/live-preview-sdk/src/rest/entities.ts
+++ b/packages/live-preview-sdk/src/rest/entities.ts
@@ -12,7 +12,7 @@ import {
 } from '../helpers/entities';
 import { ContentType, EntityReferenceMap } from '../types';
 
-type Reference = Asset | Entry | WithResourceName<Entry>;
+export type Reference = Asset | Entry | WithResourceName<Entry>;
 
 /**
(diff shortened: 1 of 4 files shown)


chrishelgert (approved):

packages/live-preview-sdk/src/rest/entities.ts
```diff
@@ -36,30 +36,27 @@ async function updateRef(
...
   // If the ID of the updateFromEntryEditor is in visitedReferences, then stop the recursion
-  if (visitedReferences.has(updateFromEntryEditor.sys.id)) {
+  if (visitedReferenceMap.has(id)) {
     debug.warn('Detected a circular reference, stopping recursion');
-    reference = entityReferenceMap.get(
-      'urn' in updateFromEntryEditor.sys
-        ? updateFromEntryEditor.sys.urn
-        : updateFromEntryEditor.sys.id
-    );
+    reference = visitedReferenceMap.get(id);
```
YvesRijckaert, line 50: Do we now have two maps, one for the entityReferenceMap and one visitedReferenceMap? Can't we just have one?

PR #383, opened 2023-10-26
fix(gql): inifite loop with circular references []

chrishelgert:
- Use the depth limit for GraphQL if no gqlParams have been provided, to prevent an endless loop of updates for circular references
- Drop VisitedReferencesMap and EntityReferenceMap in favor of the EditorEntityStore
- Move the store up on a higher level to be better accessible and increase update speed
- Move the depth handling the other way around, instead of increase the depth, decrease the remaining depth
- Additionally define a custom depth for rich text (only two levels instead of max 10)

### DIFF
--- a/packages/live-preview-sdk/src/constants.ts
+++ b/packages/live-preview-sdk/src/constants.ts
@@ -9,6 +9,7 @@ export const TOOLTIP_HEIGHT = 32;
 export const TOOLTIP_PADDING_LEFT = 5;
 
 export const MAX_DEPTH = 10;
+export const MAX_RTE_DEPTH = 2;
 
 export const LIVE_PREVIEW_EDITOR_SOURCE = 'live-preview-editor' as const;
 export const LIVE_PREVIEW_SDK_SOURCE = 'live-preview-sdk' as const;
--- a/packages/live-preview-sdk/src/rest/entities.ts
+++ b/packages/live-preview-sdk/src/rest/entities.ts
@@ -17,7 +17,7 @@ import {
   isResourceLink,
   isRichText,
 } from '../helpers/entities';
-import { ContentType, EntityReferenceMap } from '../types';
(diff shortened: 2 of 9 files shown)


packages/live-preview-sdk/src/rest/entities.ts
```diff
@@ -33,75 +26,61 @@ function getFieldName(contentType: ContentType, field: ContentType['fields'][num
...
  */
 async function updateRef(
   dataFromPreviewApp: Reference | undefined,
   updateFromEntryEditor: Reference,
   locale: string,
-  entityReferenceMap: EntityReferenceMap,
-  depth: number,
-  visitedReferenceMap: Map<string, Reference>,
-  sendMessage: SendMessage
+  maxDepth: number,
```
aodhagan-cf, line 45: why have you changed this from depth to maxDepth sorry I think I'm missing this

chrishelgert: So we can define a different _maxDepth_ for RichText

it's easier in this one, as we otherwise would need to start with `8` for rich text and check for the limit of 10

packages/live-preview-sdk/src/constants.ts
```diff
@@ -9,6 +9,7 @@ export const TOOLTIP_HEIGHT = 32;
 export const TOOLTIP_PADDING_LEFT = 5;
 
 export const MAX_DEPTH = 10;
+export const MAX_RTE_DEPTH = 2;
```
YvesRijckaert, line 12: Can you explain why we need 2 for rich text?

chrishelgert: 1 Level depth for the direct references
Assuming it's an entry it could be relevant to resolve also the asset of this entry, therefore then 2

PR #405, opened 2023-11-07
feat(inspectorMode): move inspectorMode outlines on top of the iframe []

chrishelgert:
Move the outlines and interactions of the tagged elements on top of the iframe

**Benefit:**
- better handling
- easier to add more integrations in contentful

**Refactored:**
- Move types, constants and enums to the inspectorMode directory

**Deprecates:**
- styles.css


packages/live-preview-sdk/src/styles.css
chrishelgert, line 1: This would be kind of breaking change, should we continue to ship it empty?

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -0,0 +1,171 @@
...
+    this.defaultLocale = locale;
+    this.targetOrigin = targetOrigin;
+
+    // TODO: we we need this?
+    this.onMouseOver = this.onMouseOver.bind(this);
+    this.onScroll = this.onScroll.bind(this);
+    this.handleElementInteraction = this.handleElementInteraction.bind(this);
+    this.sendAllElements = this.sendAllElements.bind(this);
+
+    // TODO: on resize do the something similar as onScroll
```
chrishelgert, line 31: remove comment

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -0,0 +1,171 @@
...
+    window.addEventListener('scroll', this.onScroll);
```
chrishelgert, line 32: should we provide a cleanup fn to remove all the listeners (maybe future improvement)

packages/live-preview-sdk/src/helpers/utils.ts
```diff
@@ -8,6 +8,7 @@ import { debug } from './debug';
  * Sends the given message to the editor
  * enhances it with the information necessary to be accepted
  */
+// FIXME: method and data doesn't need to have an overlap
```
chrishelgert, line 11: TODO: Recheck

packages/live-preview-sdk/src/styles.css
```diff
@@ -1,50 +1 @@
...
-}
-
-button.contentful-tooltip:active:hover {
-  background: rgb(0, 65, 171);
-}
-
-.contentful-inspector--active button.contentful-tooltip {
-  display: flex;
-}
+/* TODO: remove with next breaking version */
```
YvesRijckaert, line 1: Why can't we remove it now already as part of a breaking version release?

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -0,0 +1,212 @@
...
+      if (!this.isResizing) {
+        this.isScrolling = true;
+        sendMessageToEditor(InspectorModeEventMethods.RESIZE_START, {}, this.targetOrigin);
+      }
+
+      if (this.resizeTimeout) {
+        clearTimeout(this.resizeTimeout);
+      }
+
+      this.resizeTimeout = setTimeout(() => {
```
YvesRijckaert, line 149: Why are we using timeout here? for debounce / throttling?

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -0,0 +1,212 @@
...
+    });
+
+    return () => mutationObserver.disconnect();
+  }
+
+  /** Sends resize start and end event to the editor, on end it also sends the tagged elements again */
+  private bindResizeListener() {
+    const resizeObserver = new ResizeObserver(() => {
+      if (!this.isResizing) {
+        this.isScrolling = true;
```
YvesRijckaert, line 141: why are we setting isScrolling to true when you're resizing?

chrishelgert: Nice catch 🙈

packages/live-preview-sdk/src/inspectorMode/index.ts
chrishelgert: To get the information when the resizing (same for scrolling) ends and we can show the overlay in the live preview again.
During this interactions, the visuals would be not on the correct position

YvesRijckaert: Maybe we can add this as a small note? 👍

chrishelgert: should be already there now :) (same for scrolling)

packages/live-preview-sdk/src/helpers/utils.ts
```diff
@@ -13,14 +13,14 @@ export function sendMessageToEditor(
...
-  const message: MessageFromSDK = {
+  const message = {
     ...data,
     method,
     from: 'live-preview',
     source: LIVE_PREVIEW_SDK_SOURCE,
     location: window.location.href,
     version,
-  };
+  } as MessageFromSDK;
```
aodhagan-cf, line 23: out of curiosity, why?

packages/live-preview-sdk/src/messages.ts
```diff
@@ -5,6 +5,15 @@ import type { SysLink } from 'contentful-management';
 
 import type { LIVE_PREVIEW_EDITOR_SOURCE, LIVE_PREVIEW_SDK_SOURCE } from './constants';
 import { sendMessageToEditor } from './helpers';
+import {
```
aodhagan-cf, line 8: `import type` for any of these?

packages/live-preview-sdk/src/helpers/utils.ts
chrishelgert: Some overlap/confusion inside of typescript due all the different message types.
I will create a story to improve the types for the messages in general as it could produce errors

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    this.bindHoverListener = this.bindHoverListener.bind(this);
+    this.bindScrollListener = this.bindScrollListener.bind(this);
+    this.bindMutationListener = this.bindMutationListener.bind(this);
+    this.bindResizeListener = this.bindResizeListener.bind(this);
+
+    this.handleTaggedElement = this.handleTaggedElement.bind(this);
+    this.sendAllElements = this.sendAllElements.bind(this);
+
+    // Attach interaction listeners
```
aodhagan-cf, line 35: nit: maybe it's just me being super picky but I think the naming is a little confusing here as these functions are already "bound", I do understand that it's a different type of binding just as a suggestion maybe something like

```
  this.handleHoverEvents();
  this.handleScrollEvents();
  this.handleMutationObserverEvents();
  this.handleResizeEvents();
```

then the handlers are bound to the class instance 🤷🏻‍♂️

aodhagan-cf: or even `this.addHoverEventListener...` etc.

packages/live-preview-sdk/src/helpers/utils.ts
aodhagan-cf: Okay cool, thanks for explaining!

aodhagan-cf (approved):
Looks great! ⭐️

YvesRijckaert (approved):
💯 nice work!

packages/live-preview-sdk/src/inspectorMode/index.ts
chrishelgert: you want it, you get it :) 
instead of `bind` it's now `addXYZ`

PR #406, opened 2023-11-07
fix(entity-store): split into asset and entry map to prevent id collision

chrishelgert:
Entry and Asset can have the same ID, therefore we should split up the entities into two maps (one for asset, one for entry)

### DIFF
--- a/packages/visual-sdk/src/store/EditorStore.spec.ts
+++ b/packages/visual-sdk/src/store/EditorStore.spec.ts
@@ -22,6 +22,21 @@ describe('EntityStore', () => {
     expect(createStore().entities).toEqual(entities);
   });
 
+  it('prevents id collision for assets and entries', async () => {
+    const patchedAsset = { ...asset, sys: { ...asset.sys, id: '1' } };
+    const patchedEntry = { ...entry, sys: { ...entry.sys, id: '1' } };
+
+    const store = new EntityStore({ entities: [patchedAsset, patchedEntry], locale });
+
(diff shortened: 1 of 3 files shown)


packages/visual-sdk/src/store/EditorStore.spec.ts
```diff
@@ -22,6 +22,21 @@ describe('EntityStore', () => {
...
   });
 
+  it('prevents id collision for assets and entries', async () => {
+    const patchedAsset = { ...asset, sys: { ...asset.sys, id: '1' } };
+    const patchedEntry = { ...entry, sys: { ...entry.sys, id: '1' } };
+
+    const store = new EntityStore({ entities: [patchedAsset, patchedEntry], locale });
+
+    const [resolvedAsset, resolvedEntry] = await Promise.all([
+      store.fetchAsset('1'),
```
aodhagan-cf, line 32: What happens with?

```
public get entities() {
  return [...this.entryMap.values(), ...this.assetMap.values()];
}
```

aodhagan-cf: () is it possible to have duplicates? EntityStore

chrishelgert: yes there it is possible to have duplicate `sys.id`, but you also have the `sys.type` to differentiate between them, so that should be fine :)

2wce (approved):

aodhagan-cf (approved):

PR #421, opened 2023-11-14
feat: remove merge function

2wce:
This will be a WIP until the GraphQL update logic is moved

### Done

- [x] copied Rest specific update logic
- [x] removed merge function

### In progress

- [ ] copy graphql update logic

### Todo

- [ ] delete rest & graphql update logic
- [ ] update typings


packages/live-preview-sdk/src/liveUpdates.ts
```diff
@@ -1,5 +1,4 @@
-import type { Asset, Entry } from 'contentful';
-
+import { stringify } from 'flatted';
```
chrishelgert, line 1: Think we should add this as dependency instead of devDependency

packages/live-preview-sdk/src/messages.ts
```diff
@@ -164,6 +166,7 @@ export type MessageFromEditor = (
   | DebugModeEnabledMessage
   | RequestedEntitiesMessage
 ) & {
+  data: any;
```
chrishelgert, line 169: are we cleaning up the typings in another PR then?

2wce: yes, we need a couple of PRs before work on this repo & the packages one is complete

PR #436, opened 2023-11-23
chore: add next 14 revalidate example [TOL-1602]

2wce:
Added an example for live preview using next 14

- [x] instead of reloading the path, revalidate the cache for the path and update that way


examples/next-14-app-router-ssr/README.md
```diff
@@ -0,0 +1,47 @@
+# Next.js GraphQL Contentful live preview SDK example
+
+This is an example project that demonstrates how to use the `@contentful/live-preview` SDK with a Next.js application that runs only on the server.
```
YvesRijckaert, line 3: Can we add some more information how this example differs from the other Next.js examples? Right now it is not very clear to me what the benefits are of this one vs the other examples

Maybe even adding that to the general READ me in the Next js section

chrishelgert:
Could you give is a quick intro on how it works in the PH? :)

chrishelgert (approved):

PR #450, opened 2023-11-27
fix(validation): warn about invalid query usage on live updates subscription []

chrishelgert:
- Warn if the GraphQL query is provided in the wrong format
- Warn if the query is provided on a REST setup
- Remove the invalid param to prevent breaking apps

### DIFF
--- a/packages/live-preview-sdk/src/__tests__/validation.spec.ts
+++ b/packages/live-preview-sdk/src/__tests__/validation.spec.ts
@@ -1,177 +1,330 @@
 import { describe, it, expect, vi, afterEach } from 'vitest';
 
 import { debug } from '../helpers';
-import { validateDataForLiveUpdates } from '../helpers/validation';
+import { validateLiveUpdatesConfiguration } from '../helpers/validation';
+import gql from 'graphql-tag';
 
 vi.mock('../helpers/debug');
 
-describe('validateDataForLiveUpdates', () => {
+describe('validateLiveUpdatesConfiguration', () => {
+  const callback = vi.fn();
+
   afterEach(() => {
     vi.clearAllMocks();
   });
 
-  describe('REST', () => {
-    it('detects REST structure on the top level', () => {
-      const result = validateDataForLiveUpdates({
-        sys: { id: '123' },
-        fields: {
-          title: { 'en-US': 'Hello World' },
-        },
-      });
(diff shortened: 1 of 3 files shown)


packages/live-preview-sdk/src/__tests__/validation.spec.ts
```diff
@@ -1,177 +1,363 @@
...
+          isGQL: true,
+          isREST: false,
+          sysId: '123',
+          isValid: true,
+          config,
+        });
       });
 
-      expect(debug.error).not.toHaveBeenCalled();
+      it('detects REST structure in a nested structure', () => {
```
aodhagan-cf, line 174: Is this not GQL?

packages/live-preview-sdk/src/helpers/validation.ts
```diff
@@ -22,10 +33,34 @@ function validation(d: Argument): { isGQL: boolean; sysId: string | null; isREST
...
-    return validation(Object.values(d) as Argument);
+    return validateData(Object.values(d) as Argument, maxDepth - 1);
   }
 }
 
+function validatedConfig(originalConfig: ContentfulSubscribeConfig, isREST: boolean) {
+  const config = { ...originalConfig };
+
+  if (config.query) {
+    if (typeof config.query !== 'object') {
```
aodhagan-cf, line 44: would we consider parsing it in future?

aodhagan-cf (approved):
LGTM

packages/live-preview-sdk/src/helpers/validation.ts
chrishelgert: that would increase the bundle size of our package as we need graphql-tag as a dependency (at least peerDependency)
We could try with optional dependency maybe, can give it a try tomorrow

aodhagan-cf: this is kind of interesting: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

chrishelgert: graphql-tag is 2kb, so I imported it directly now
Otherwise we would rework the whole subscriptions to be async :(

2wce (approved):

PR #452, opened 2023-12-01
chore: adding resizing handler for entry updates [TOL-1644]

aodhagan-cf:
Adding resize handling for entry saves and updates

### DIFF
--- a/packages/live-preview-sdk/src/inspectorMode/index.ts
+++ b/packages/live-preview-sdk/src/inspectorMode/index.ts
@@ -134,11 +134,6 @@ export class InspectorMode {
       subtree: true,
     });
 
-    // mutationObserver.observe(description, {
-    //   attributes: true,characterData: true, subtree: true, childList: true
-
-    //   })
-
     return () => mutationObserver.disconnect();
   }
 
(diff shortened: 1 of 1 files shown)


packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -46,9 +46,39 @@ export class InspectorMode {
       if (isInspectorActive) {
         this.sendAllElements();
       }
+    } else if (
+      data.method === LivePreviewPostMessageMethods.ENTRY_UPDATED ||
+      data.method === LivePreviewPostMessageMethods.ENTRY_SAVED
+    ) {
+      // for entry updates we need to wait a bit to make sure the DOM is updated
+      // then send the tagged elements again so that outlines are correct
+      this.handleResizing(250);
```
chrishelgert, line 55: Did you try adjusting the mutationObserver?
Maybe there would be something to optimize then we would not need this arbitrary number :)

aodhagan-cf: It wasn't working for me before but since passing `characterData: true` it works

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -202,15 +207,36 @@ export class InspectorMode {
...
+        ],
+        childList: true,
+        subtree: true,
+        characterData: true,
+      });
+    });
+
+    sendTaggedElementsMessage();
+
+    return () => mutationObserver.disconnect();
```
chrishelgert, line 240: I think this cleanup function will never be called :(

aodhagan-cf: Ah right because this isn't added as a event listener callback...

chrishelgert (approved):
LGTM

PR #473, opened 2023-12-14
feat: implement CSM transformation to encoded strings for GraphQL [TOL-1706]

YvesRijckaert:
Transform Content Source Maps (CSM) from the GraphQL API into hidden encoded strings. Only fields that are not evaluated as URL or ISO Date.

### DIFF
--- a/packages/live-preview-sdk/package.json
+++ b/packages/live-preview-sdk/package.json
@@ -51,7 +51,6 @@
   "dependencies": {
     "@contentful/rich-text-types": "^16.2.0",
     "@contentful/visual-sdk": "^1.0.0-alpha.34",
-    "@types/json-pointer": "^1.0.34",
     "@vercel/stega": "^0.1.0",
     "graphql-tag": "^2.12.6",
     "json-pointer": "^0.6.2",
@@ -59,6 +58,7 @@
   },
   "devDependencies": {
     "@testing-library/react": "14.1.2",
+    "@types/json-pointer": "^1.0.34",
     "@types/lodash.isequal": "^4.5.6",
     "@types/node": "^20.1.0",
     "@types/react": "^18.0.27",
--- a/packages/live-preview-sdk/package.json
+++ b/packages/live-preview-sdk/package.json
@@ -51,7 +51,10 @@
   "dependencies": {
     "@contentful/rich-text-types": "^16.2.0",
     "@contentful/visual-sdk": "^1.0.0-alpha.34",
+    "@types/json-pointer": "^1.0.34",
+    "@vercel/stega": "^0.1.0",
     "graphql-tag": "^2.12.6",
+    "json-pointer": "^0.6.2",
     "lodash.isequal": "^4.5.0"
   },
   "devDependencies": {
--- a/packages/live-preview-sdk/src/__tests__/csm.spec.ts
+++ b/packages/live-preview-sdk/src/__tests__/csm.spec.ts
@@ -2,20 +2,8 @@ import { describe, test, expect } from 'vitest';
 import { encodeSourceMap } from '../csm';
 import { vercelStegaDecode } from '@vercel/stega';
 
-type Mappings = {
-  [key: string]:
-    | {
-        origin: string;
-        href: string;
-      }
-    | {
-        [nestedKey: string]: {
-          origin: string;
-          href: string;
-        };
-      }
-    | undefined;
-};
+type Mapping = { origin: string; href: string };
+type Mappings = Record<string, Mapping | Record<string, Mapping> | undefined>;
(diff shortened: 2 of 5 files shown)


packages/live-preview-sdk/package.json
```diff
@@ -51,7 +51,10 @@
   "dependencies": {
     "@contentful/rich-text-types": "^16.2.0",
     "@contentful/visual-sdk": "^1.0.0-alpha.34",
+    "@types/json-pointer": "^1.0.34",
```
chrishelgert, line 54: @types should be only devDependencies I think

packages/live-preview-sdk/src/csm/types.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface EntitySource {
+  space: number;
+  environment: number;
+  id: string;
+}
+
+type Mappings = Record<string, { source: Source }>;
+
+interface IContentSourceMaps {
```
chrishelgert, line 14: Why do we have the `I` prefix here? 🤔

packages/live-preview-sdk/src/csm/index.ts
```diff
@@ -0,0 +1,74 @@
+import { vercelStegaEncode } from '@vercel/stega';
+import { debug } from '../helpers';
+import jsonPointer from 'json-pointer';
+import { EntitySource, GraphQLResponse, Source } from './types';
+
+export class ContentSourceMaps {
```
chrishelgert, line 6: Wondering if we need the class as it has no constructor, internal data or something like that and only exposes one public function

We could also use only functions and only export the relevant one

packages/live-preview-sdk/src/csm/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    const entity = isEntry ? entries[source.entry] : assets[source.asset];
+    if (!entity) return null;
+
+    const space = spaces[entity.space];
+    const environment = environments[entity.environment];
+    const entityId = entity.id;
+    const field = fields[source.field];
+    const locale = locales[source.locale];
+
+    const basePath = `https://app.contentful.com/spaces/${space}/environments/${environment}`;
```
chrishelgert, line 36: Does this need to handle eu residency?

packages/live-preview-sdk/src/csm/types.ts
YvesRijckaert: Ahh, this was in the same file as the ContentSourceMaps class, so I needed to rename it (I for interface). But since I moved it to its own types file we can remove the I

packages/live-preview-sdk/src/csm/index.ts
YvesRijckaert: Sure that's good for me, I went with a class since the other files are also using that approach

packages/live-preview-sdk/src/csm/index.ts
YvesRijckaert: Good call! 🇪🇺
Wonder how I would handle that. I guess somehow relying on the targetOrigin of ContentfulLivePreviewInitConfig? Or do we accept a second parameter for the encodeSourceMap function that has the targetOrigin?

packages/live-preview-sdk/src/__tests__/csm.spec.ts
```diff
@@ -0,0 +1,365 @@
+import { describe, test, expect } from 'vitest';
+import { encodeSourceMap } from '../csm';
+import { vercelStegaDecode } from '@vercel/stega';
+
+type Mappings = {
+  [key: string]:
+    | {
```
chrishelgert, line 7: NIT: extracting the origin + href object into a separate type could make it more readable
Addition use `Record`

```
// Demo code
type Mapping = { origin: string; href: string; } 
type Mappings = Record<string, Mapping | Record<string, Mapping> | undefined>
```

packages/live-preview-sdk/src/csm/index.ts
```diff
@@ -0,0 +1,86 @@
+import { vercelStegaEncode } from '@vercel/stega';
+import { debug } from '../helpers';
+import jsonPointer from 'json-pointer';
+import { EntitySource, GraphQLResponse, Source } from './types';
+
+const isUrlOrIsoDate = (value: string) => {
```
chrishelgert, line 6: ❓ couldn't we solve that later on with the content type?

chrishelgert: ah but also false positives as it is only appearance and still symbol in the background right?
then nvm

chrishelgert (approved):
LGTM

PR #482, opened 2023-12-21
fix: cleanup request cache after timeout [SPA-1711]

Chaoste:
Currently, a timed out request is kept in the request cache. So when we retry it, we directly get the same error again.


YvesRijckaert (approved):
lgtm!

PR #509, opened 2024-01-17
feat(live-updates): send unsubscribed message to cleanup subscriptions on the counter part []

chrishelgert:


packages/live-preview-sdk/src/liveUpdates.ts
```diff
@@ -142,17 +143,22 @@ export class LiveUpdates {
...
       config: stringify(config),
+    };
+
+    this.sendMessage(LivePreviewPostMessageMethods.SUBSCRIBED, {
+      action: LivePreviewPostMessageMethods.SUBSCRIBED,
+      ...message,
     } as SubscribedMessage);
 
     return () => {
+      this.sendMessage(LivePreviewPostMessageMethods.UNSUBSCRIBED, message);
```
chrishelgert, line 161: We can then pick this message up on the user_interface and remove it from the subscription map there

2wce (approved):

PR #510, opened 2024-01-17
chore(visual-sdk): remove as it's no longer needed []

chrishelgert:
The visual-sdk is no longer needed for the canary version and also no longer shared internally,


colomolo:
I've removed it from master a while ago. We don't need it _at all_?

chrishelgert:
Yes not needed anymore, more details in slack

colomolo (approved):

aodhagan-cf (approved):

PR #522, opened 2024-01-22
refactor: small clean up and adding tests

z0al:
Small cleanup:
- Suppress expected errors from test log
- Added tests for `getAllTaggedElements` and added checking for proper attributes there.
- Used `TreeWalker` for a faster & more flexible document scanning. 
- Random fixes

I opened this to keep the follow up PR small. There should be **no behavioral** change with this PR.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -31,8 +31,15 @@ module.exports = {
       },
     ],
     'react/react-in-jsx-scope': 'off',
-    '@typescript-eslint/no-empty-function': 'off',
   },
+  overrides: [
+    {
+      files: ['*.test.ts(x)', '*.spec.tsx(x)'],
+      rules: {
+        '@typescript-eslint/no-empty-function': 'off',
+      },
+    },
+  ],
   settings: {
     react: { version: 'detect' },
   },
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -31,6 +31,7 @@ module.exports = {
       },
     ],
     'react/react-in-jsx-scope': 'off',
+    '@typescript-eslint/no-empty-function': 'off',
   },
   settings: {
     react: { version: 'detect' },
--- a/packages/live-preview-sdk/src/inspectorMode/__tests__/getAllTaggedElements.test.ts
+++ b/packages/live-preview-sdk/src/inspectorMode/__tests__/getAllTaggedElements.test.ts
@@ -55,27 +55,30 @@ describe('getAllTaggedElements', () => {
     it('should ignore elements without field id', () => {
       const dom = html(`
 		<div>
-		  <!-- Entries -->
-		  <div ${dataEntry}="entry-1"></div>
+		  <!-- Keep -->
+		  <div id="entry" ${dataEntry}="entry-id" ${dataField}="field-id"></div>
 
+		  <!-- Ignore -->
+		  <div ${dataEntry}="entry-1"></div>
 		  <div ${dataEntry}="entry-2" ${dataLocale}="locale-2"></div>
 
-		  <!-- Assets -->
 		  <div ${dataAsset}="asset-1"></div>
-
 		  <div ${dataAsset}="asset-2" ${dataLocale}="locale-2"></div>
 		</div>
 	  `);
 
       const elements = getAllTaggedElements(dom);
 
-      expect(elements).toEqual([]);
+      expect(elements).toEqual([dom.getElementById('entry')]);
     });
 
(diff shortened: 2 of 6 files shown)


packages/live-preview-sdk/src/inspectorMode/__tests__/getAllTaggedElements.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    it('should ignore elements without field id', () => {
+      const dom = html(`
+		<div>
+		  <!-- Entries -->
+		  <div ${dataEntry}="entry-1"></div>
+
+		  <div ${dataEntry}="entry-2" ${dataLocale}="locale-2"></div>
+
```
aodhagan-cf, line 62: nit: maybe have some with the field id so show that it's not just ignoring everything?

.eslintrc.cjs
```diff
@@ -31,6 +31,7 @@ module.exports = {
       },
     ],
     'react/react-in-jsx-scope': 'off',
+    '@typescript-eslint/no-empty-function': 'off',
```
chrishelgert, line 34: Would you mind disabling this only for the test files?

```
"overrides": [
    {
      "files": ["src/**/*.test.ts(x)", "src/**/*.spec.ts(x)"],
      "rules": {
        '@typescript-eslint/no-empty-function': 'off',
      }
    }
  ]
```

Not 100% sure about the syntax but should be something like this

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -1,40 +1,77 @@
 import { InspectorModeAttributes, InspectorModeDataAttributes } from './types';
 
+const isTaggedElement = (node: Node): boolean => {
+  // is Element?
```
aodhagan-cf, line 4: I'm not sure this comment helps we understand more what is going on

aodhagan-cf (approved):
LGTM, just a few small comments

packages/live-preview-sdk/src/inspectorMode/utils.ts
z0al: At first I was using `node.nodeType === 1` before discovering the `Node.ELEMENT_NODE`

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -21,18 +23,7 @@ export class InspectorMode {
...
-
-    this.addHoverListener = this.addHoverListener.bind(this);
-    this.addScrollListener = this.addScrollListener.bind(this);
-    this.addMutationListener = this.addMutationListener.bind(this);
-    this.addResizeListener = this.addResizeListener.bind(this);
-
-    this.handleTaggedElement = this.handleTaggedElement.bind(this);
-    this.sendAllElements = this.sendAllElements.bind(this);
-
+  constructor(private options: InspectorModeOptions) {
```
chrishelgert, line 26: TIL
didn't know that this is working

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -1,40 +1,77 @@
...
-  ];
-  return allElements;
+export function getAllTaggedElements(root = window.document): Element[] {
+  // The fastest way to look up & iterate over DOM. Ref:
+  // https://stackoverflow.com/a/2579869
+  //
+  // Terminology:
+  // FILTER_SKIP: Skip the current node
+  // FILTER_REJECT: Skip the current node and all its children
+  const walker = document.createTreeWalker(root, NodeFilter.SHOW_ALL, (node) => {
```
aodhagan-cf, line 64: Maybe we could even call this an ent haha https://lotr.fandom.com/wiki/Ents

packages/live-preview-sdk/src/inspectorMode/utils.ts
z0al: Yeah, sure I will remove the comment. I was just explaining why it was there

aodhagan-cf: Thanks mate!

.eslintrc.cjs
z0al: Fixed

packages/live-preview-sdk/src/inspectorMode/__tests__/getAllTaggedElements.test.ts
z0al: Fixed

PR #523, opened 2024-01-25
feat: recognise auto-tagged elements [TOL-1711]

z0al:
- Detect hidden strings in a page and convert it to proper data attributes.
  - Note that this doesn't care about whether a parent element is manually tagged or not.
- Adds a flag to ignore existing manual attributes for testing purposes.

### DIFF
--- a/packages/live-preview-sdk/src/__tests__/csm.spec.ts
+++ b/packages/live-preview-sdk/src/__tests__/csm.spec.ts
@@ -2,9 +2,9 @@ import { vercelStegaDecode } from '@vercel/stega';
 import { describe, test, expect } from 'vitest';
 
 import { encodeSourceMap } from '../csm';
+import { SourceMapMetadata } from '../csm/encode';
 
-type Mapping = { origin: string; href: string };
-type Mappings = Record<string, Mapping | Record<string, Mapping> | undefined>;
+type Mappings = Record<string, SourceMapMetadata | Record<string, SourceMapMetadata> | undefined>;
 
 type EncodedResponse =
   | {
@@ -77,10 +77,26 @@ describe('Content Source Maps', () => {
         title: {
           origin: 'contentful.com',
           href: 'https://app.contentful.com/spaces/foo/environments/master/entries/a1b2c3/?focusedField=title&focusedLocale=en-US',
+          cf: {
+            space: 'foo',
+            environment: 'master',
+            field: 'title',
+            locale: 'en-US',
+            entity: 'a1b2c3',
+            entityType: 'Entry',
+          },
         },
         subtitle: {
           origin: 'contentful.com',
           href: 'https://app.contentful.com/spaces/foo/environments/master/entries/a1b2c3/?focusedField=subtitle&focusedLocale=en-US',
+          cf: {
+            space: 'foo',
+            environment: 'master',
+            field: 'subtitle',
+            locale: 'en-US',
+            entity: 'a1b2c3',
+            entityType: 'Entry',
+          },
         },
       });
     });
@@ -129,10 +145,26 @@ describe('Content Source Maps', () => {
         title: {
           origin: 'contentful.com',
           href: 'https://app.eu.contentful.com/spaces/foo/environments/master/entries/a1b2c3/?focusedField=title&focusedLocale=en-US',
+          cf: {
+            space: 'foo',
+            environment: 'master',
+            field: 'title',
+            locale: 'en-US',
+            entity: 'a1b2c3',
+            entityType: 'Entry',
+          },
         },
         subtitle: {
           origin: 'contentful.com',
           href: 'https://app.eu.contentful.com/spaces/foo/environments/master/entries/a1b2c3/?focusedField=subtitle&focusedLocale=en-US',
+          cf: {
+            space: 'foo',
+            environment: 'master',
+            field: 'subtitle',
+            locale: 'en-US',
+            entity: 'a1b2c3',
+            entityType: 'Entry',
+          },
         },
       });
     });
@@ -199,18 +231,42 @@ describe('Content Source Maps', () => {
           title: {
             origin: 'contentful.com',
             href: 'https://app.contentful.com/spaces/foo/environments/master/entries/a1b2c3/?focusedField=title&focusedLocale=en-US',
+            cf: {
+              space: 'foo',
+              environment: 'master',
+              field: 'title',
+              locale: 'en-US',
+              entity: 'a1b2c3',
+              entityType: 'Entry',
+            },
           },
         },
         1: {
           title: {
             origin: 'contentful.com',
             href: 'https://app.contentful.com/spaces/foo/environments/master/entries/d4e5f6/?focusedField=title&focusedLocale=en-US',
+            cf: {
+              space: 'foo',
+              environment: 'master',
+              field: 'title',
+              locale: 'en-US',
+              entity: 'd4e5f6',
+              entityType: 'Entry',
+            },
           },
         },
         2: {
           title: {
             origin: 'contentful.com',
             href: 'https://app.contentful.com/spaces/foo/environments/master/entries/g7h8i9/?focusedField=title&focusedLocale=en-US',
+            cf: {
+              space: 'foo',
+              environment: 'master',
+              field: 'title',
+              locale: 'en-US',
+              entity: 'g7h8i9',
+              entityType: 'Entry',
+            },
           },
         },
       });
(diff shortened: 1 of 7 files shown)


packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -53,21 +58,87 @@ export function getInspectorModeAttributes(
...
+    // Handle Encoded strings
+    const { href } = (vercelStegaDecode(node.textContent ?? '') ?? {}) as Record<string, string>;
+    const el = node.parentElement;
+
+    if (!el) {
+      console.log('skipping no parent');
+      continue;
+    }
+
+    // FIXME: maybe we should encode these attributes as separate attributes
```
z0al, line 111: any objection for adding these info as separate attributes?

chrishelgert: You mean not parsing them from the URL and having them directly as properties?

```
const { spaceId, environmentId, entryId, assetId, fieldId, localeCode } = vercelStegaDecode(node.textContent).data
```

Only con I would see is that the hidden data would be even more payload and it could affect the loading time (more) than having the information only once

z0al: > You mean not parsing them from the URL and having them directly as properties?

Exactly.

chrishelgert: As mentioned the only con I would see is regarding performance
But I would be also in favor of having the data directly without the additional need for parsing, makes it more clearer and easier to understand/build on

z0al: I agree. But also, with the payload size it may not be big of an issue since it's only enabled for preview deployments.

packages/live-preview-sdk/src/__tests__/csm.spec.ts
```diff
@@ -2,9 +2,9 @@ import { vercelStegaDecode } from '@vercel/stega';
 import { describe, test, expect } from 'vitest';
 
 import { encodeSourceMap } from '../csm';
+import { SourceMapMetadata } from '../csm/encode';
```
aodhagan-cf, line 5: import type maybe?

packages/live-preview-sdk/src/__tests__/csm.spec.ts
```diff
@@ -77,10 +77,26 @@ describe('Content Source Maps', () => {
         title: {
           origin: 'contentful.com',
           href: 'https://app.contentful.com/spaces/foo/environments/master/entries/a1b2c3/?focusedField=title&focusedLocale=en-US',
+          cf: {
```
aodhagan-cf, line 80: Sorry but what does `cf` mean here? is there a more meaningful property name you could give it?

packages/live-preview-sdk/src/csm/index.ts
```diff
@@ -58,24 +48,37 @@ export const encodeSourceMap = (
...
-      fields,
-      locales,
-      targetOrigin
-    );
+
+    const entity = 'entry' in source ? entries[source.entry] : assets[source.asset];
+    const entityType = 'entry' in source ? 'Entry' : 'Asset';
+
+    if (!entity) {
+      graphqlResponse;
```
chrishelgert, line 56: ❓ what is this line doing? 🤔

packages/live-preview-sdk/src/csm/index.ts
```diff
@@ -15,29 +15,19 @@ const isUrlOrIsoDate = (value: string) => {
...
-  fields: string[],
-  locales: string[],
+  entityId: string,
+  entityType: 'Entry' | 'Asset',
+  space: string,
+  environment: string,
+  field: string,
+  locale: string,
   targetOrigin?: 'https://app.contentful.com' | 'https://app.eu.contentful.com'
 ): string | null => {
```
chrishelgert, line 25: this no longer returns null I think

packages/live-preview-sdk/src/csm/encode.ts
```diff
@@ -0,0 +1,22 @@
+import { vercelStegaDecode, vercelStegaEncode } from '@vercel/stega';
+
+export type SourceMapMetadata = {
+  origin: string;
+  href: string;
+  cf: {
+    space: string;
+    environment: string;
```
aodhagan-cf, line 8: can `environment` be undefined?

packages/live-preview-sdk/src/csm/index.ts
```diff
@@ -58,24 +48,37 @@ export const encodeSourceMap = (
...
+
+    const space = spaces[entity.space];
+    const environment = environments[entity.environment];
+    const entityId = entity.id;
+    const field = fields[source.field];
+    const locale = locales[source.locale];
+
+    const href = getHref(entityId, entityType, space, environment, field, locale, targetOrigin);
 
     if (href && jsonPointer.has(data, pointer)) {
```
chrishelgert, line 67: as the href can no longer be null, we can remove it from the `if`

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -53,21 +58,66 @@ export function getInspectorModeAttributes(
...
+      continue;
+    }
+
+    if (cf.entityType === 'Entry') {
+      el.setAttribute(InspectorModeDataAttributes.ENTRY_ID, cf.entity);
+    } else {
+      el.setAttribute(InspectorModeDataAttributes.ASSET_ID, cf.entity);
+    }
+
+    // TODO: add space/env ids to properly handle cross-space content
```
chrishelgert, line 116: If it's no bigger change, could we send them directly, even if we don't handle them yet in the live-preview?

packages/live-preview-sdk/src/index.ts
```diff
@@ -120,7 +124,11 @@ export class ContentfulLivePreview {
 
       // setup the live preview plugins (inspectorMode and liveUpdates)
       if (this.inspectorModeEnabled) {
-        this.inspectorMode = new InspectorMode({ locale, targetOrigin: this.targetOrigin });
+        this.inspectorMode = new InspectorMode({
```
chrishelgert, line 127: Not in this one, but do we have already a ticket to update documentation and create an example for auto-tagging?

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -53,21 +58,66 @@ export function getInspectorModeAttributes(
...
-    elements.push(walker.currentNode as Element);
+    const node = walker.currentNode;
+
+    if (node.nodeType === Node.ELEMENT_NODE) {
+      elements.push(walker.currentNode as Element);
+      continue;
+    }
+
+    // Handle Encoded strings
+    const { cf } = decode(node.textContent ?? '') as SourceMapMetadata;
```
aodhagan-cf, line 103: > as SourceMapMetadata

Is this not enforced by the return type of the function?

Seems unnecessary to decode an empty string
```
if (node.textContent) {
  const { cf } = decode(node.textContent)
  ... set attributes
}
```

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -53,21 +58,66 @@ export function getInspectorModeAttributes(
...
+    const el = node.parentElement;
+
+    if (!el) {
+      continue;
+    }
+
+    if (cf.entityType === 'Entry') {
+      el.setAttribute(InspectorModeDataAttributes.ENTRY_ID, cf.entity);
+    } else {
```
aodhagan-cf, line 112: What if we don't have an entityType?

z0al: Things will break :). But the same is true for other attributes and it feels cumbersome to check each attribute. TS should handle that for the most part.

packages/live-preview-sdk/src/index.ts
z0al: https://contentful.atlassian.net/browse/TOL-1784

packages/live-preview-sdk/src/csm/encode.ts
z0al: Backend always returns it and we should always set it unless a bug happened somewhere.

packages/live-preview-sdk/src/csm/index.ts
z0al: 🙈

packages/live-preview-sdk/src/__tests__/csm.spec.ts
z0al: contentful?

packages/live-preview-sdk/src/inspectorMode/utils.ts
z0al: > Is this not enforced by the return type of the function?

It can be undefined with types alone. I'm forcing it it because we already ensure it exists in the preview code block

packages/live-preview-sdk/src/__tests__/csm.spec.ts
z0al: Funnily, I feel like in this repo we seem to overly express attribute names e.g. `InspectorModeDataAttributes`. in here I went to the other extreme mostly to save a few bytes but I guess that should have been applied consistently.

packages/live-preview-sdk/src/__tests__/csm.spec.ts
z0al: It's a test file :). But fixed.

aodhagan-cf (approved):
LGTM

PR #527, opened 2024-01-26
chore: add catalog info file []

chrishelgert:


catalog-info.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
```
Chaoste, line 2: FYI I used in our repo the `kind: System` to describe the monorepo and add then create a catalog file per package. This approach I copied from `experience-packages`. Might make sense to use it here if you plan to have more packages in here that should show up in backstage.

chrishelgert: Thanks updated to that :)

Chaoste (approved):

PR #552, opened 2024-02-05
feat: export csm

YvesRijckaert:
Export CSM function

### DIFF
--- a/packages/live-preview-sdk/src/index.ts
+++ b/packages/live-preview-sdk/src/index.ts
@@ -48,6 +48,10 @@ export interface ContentfulLivePreviewInitConfig {
    * Can be `https://app.contentful.com` or `https://app.eu.contentful.com`
    */
   targetOrigin?: string | string[];
+
+  experimental?: {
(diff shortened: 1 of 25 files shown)


packages/live-preview-sdk/src/index.ts
```diff
@@ -3,6 +3,8 @@ import './styles.css';
 import { type DocumentNode } from 'graphql';
 
 import { version } from '../package.json';
+import { encodeSourceMap as CSM } from './csm';
```
aodhagan-cf, line 6: why the renaming? I think encodeSourceMap makes sense no? Maybe rename to `encodeCSM`? I think we are like to export more functions from the `./csm` file in future

aodhagan-cf (approved):
Looks good

packages/live-preview-sdk/src/index.ts
YvesRijckaert: renaming wasn't necessary you're right 👍

PR #553, opened 2024-02-05
fix: adding rich text csm encoding for text nodes [TOL-1788]

aodhagan-cf:
When we are encoding the rich text response we
- find all the text nodes
- remove rich text pointer
- add a pointer for all text nodes
- add encoded value to the start of each of the text nodes

We also implicitly ignore null rich text values


packages/live-preview-sdk/src/__tests__/csm.spec.ts
```diff
@@ -361,6 +362,147 @@ describe('Content Source Maps', () => {
       });
     });
 
+    test('it should ignore null rich text values', () => {
```
YvesRijckaert, line 365: Do we also need to test if it ignores null values for other field types?

packages/live-preview-sdk/src/csm/richText.ts
```diff
@@ -0,0 +1,45 @@
+import jsonPointer from 'json-pointer';
+
+import { Mappings } from './types';
+
+export const isRichTextValue = (value: unknown): boolean =>
+  !!(value && typeof value === 'object' && 'nodeType' in value && value.nodeType);
```
MayaGillilan, line 6: We probably don't need to check if `value.nodeType` is truthy actually

PR #554, opened 2024-02-06
fix: adding rich text csm encoding for text nodes [TOL-1788]

aodhagan-cf:
When we are encoding the rich text response we

- find all the text nodes
- remove rich text pointer
- add a pointer for all text nodes
- add encoded value to the start of each of the text nodes
- We also implicitly ignore null rich text values


packages/live-preview-sdk/src/__tests__/csm.spec.ts
```diff
@@ -101,6 +102,47 @@ describe('Content Source Maps', () => {
...
+                  field: 1,
+                  locale: 0,
+                },
+              },
+            },
+          },
+        },
+      };
+      const encodedGraphQLResponse = encodeSourceMap(graphQLResponse);
+      expect(encodedGraphQLResponse.data.post.title).toBe(null);
```
chrishelgert, line 142: NIT there is a `toBeNull()` function in vitest and jest :)

chrishelgert (approved):

packages/live-preview-sdk/src/index.ts
```diff
@@ -317,3 +317,4 @@ export class ContentfulLivePreview {
 
 export { LIVE_PREVIEW_EDITOR_SOURCE, LIVE_PREVIEW_SDK_SOURCE } from './constants';
 export * from './messages';
+export * from './csm';
```
YvesRijckaert, line 320: do we export it differently now?

aodhagan-cf: I've removed this sorry should have double-checked this! thanks for spotting

PR #570, opened 2024-02-13
docs: add better Nextjs examples

YvesRijckaert:
Add better Next.js 14 examples (App router + Pages router)
- The App router example uses the GraphQL API and has logic for On-Demand Revalidation
- The Pages router example uses the REST API. I did not include On-Demand Revalidation logic here since Contentful.js is not using fetch but axios, so we can't pass along the Next.js specific header to revalidate specific content
- Both examples now include a script to import the content model, to make the setup easier
- Both examples have updated README's to make it more clear


chrishelgert (approved):
LGTM

PR #572, opened 2024-02-16
refactor: update-sysId-to-sysIds [TOL-1601]

2wce:
## Description 
 As part of the cleanup task for the stability topic we want to keep track of the multiple sys ids that a subscription listens for. This PR updates the validation functions & the tests to supports sysIds an array instead of a string.

## Warning
This should be merged together with [this web app PR](https://github.com/contentful/user_interface/pull/19630)


packages/live-preview-sdk/src/helpers/validation.ts
```diff
@@ -108,17 +111,17 @@ type ValidationResult = (
 export function validateLiveUpdatesConfiguration(
```
chrishelgert, line 111: We have the same function on live-preview-packages, should we consolidate them?

2wce: I'm removing it from the SDK & I've updated the implementation in  [live-preview-packages](https://github.com/contentful/live-preview-packages/pull/134)

packages/live-preview-sdk/package.json
```diff
@@ -44,7 +44,7 @@
   },
   "homepage": "https://github.com/contentful/live-preview#readme",
   "dependencies": {
-    "@contentful/live-preview-updates": "0.1.0-alpha.21",
+    "@contentful/live-preview-updates": "0.1.0-alpha.22",
```
chrishelgert, line 47: we can't use that one here, as it's an internal package

chrishelgert: let's do the validation only on the web app and if the validation fails there, we can send a message back to the SDK

2wce: yeah i can see that it breaks the CI as well,  we'll have to duplicate the validation logic here until we migrate live updates into the web app then

chrishelgert: No need to duplicate I think,
the logic is already in the web app, if the validation fails, we just send a SUBSCRIPTION_FAILED message back to the LP SDK and log an error with `debug`

chrishelgert: - Send subscription to the web app
- Web app validates the data/config with @contentful/live-preview-updates
- If successful -> all good live updates should be working
- If failed -> send post message with subscription failed back to the SDK (and maybe the reason)

2wce: great idea, I've created a follow up [ticket](https://contentful.atlassian.net/browse/TOL-1872?atlOrigin=eyJpIjoiNDY2ZjMyZGY0NzMyNGFiMTg1ODY1MjgwOGM3MzQ0OTkiLCJwIjoiaiJ9) to remove the duplication so that this PR's scope doesn't keep increasing

packages/live-preview-sdk/src/helpers/validation.ts
```diff
@@ -107,17 +167,17 @@ type ValidationResult = (
...
```
YvesRijckaert, line 167: Let's add a @TODO comment that we want to remove this duplication?

YvesRijckaert (approved):
LGTM

PR #584, opened 2024-02-20
chore: update build config to fix cjs issues in LP []

2wce:
# Issue
Our exports weren't correctly configured so as a result when we imported from the csm package in LP we would get an error which didn't break the CI. This was now affecting the stability cleanup PR

# Solution
- We have set up the build & export config to match the LP SDK
- we have also added a fix for handling graqhpl queries are passed as strings

### DIFF
--- a/packages/content-source-maps/package.json
+++ b/packages/content-source-maps/package.json
@@ -39,12 +39,9 @@
   },
   "dependencies": {
     "@vercel/stega": "^0.1.0",
-    "json-pointer": "^0.6.2",
-    "vitest": "^1.2.2"
+    "json-pointer": "^0.6.2"
   },
   "devDependencies": {
-    "chokidar": "3.6.0",
-    "typescript": "5.1.6",
-    "vite-plugin-dts": "2.3.0"
+    "chokidar": "3.6.0"
   }
 }
--- a/packages/content-source-maps/package.json
+++ b/packages/content-source-maps/package.json
@@ -4,14 +4,15 @@
   "version": "0.2.0-alpha.9",
   "description": "Base SDK for building Content Source Maps with Contentful",
   "license": "MIT",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
-  "types": "dist/types/index.d.ts",
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
+  "typings": "./dist/index.d.ts",
+  "type": "module",
   "exports": {
     ".": {
-      "import": "./dist/esm/index.js",
-      "require": "./dist/cjs/index.js",
-      "types": "./dist/types/index.d.ts"
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs",
+      "types": "./dist/index.d.ts"
     },
     "./package.json": "./package.json"
   },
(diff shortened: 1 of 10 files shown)


2wce:
question @contentful/team-tolkien This fix requires the csm to be fixed first otherwise the SDK fails because it's using the broken version which the PR aims to fix. is it fine to temporarily remove CSM usage in the SDK then add it back once this is merged? or any other alternative

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -58,89 +58,92 @@ export function getInspectorModeAttributes(
 /**
  * Query the document for all tagged elements
  */
-export function getAllTaggedElements(root = window.document, ignoreManual?: boolean): Element[] {
+export function getAllTaggedElements(_root = window.document, _ignoreManual?: boolean): Element[] {
```
chrishelgert, line 61: This would also break the manual tagging

2wce:
> Question if I understand it correctly: We fix the build (this PR) Then we directly open a new PR to enable the CSM code again?

yes @chrishelgert the follow up PR will be using the new release for csm so it should work

packages/content-source-maps/package.json
```diff
@@ -47,11 +43,8 @@
...
   "devDependencies": {
-    "@swc/cli": "0.1.63",
-    "@swc/core": "1.3.100",
-    "@swc/jest": "0.2.29",
-    "@types/jest": "29.5.11",
-    "jest": "29.7.0",
-    "typescript": "5.1.6"
+    "chokidar": "3.6.0",
+    "typescript": "5.1.6",
+    "vite-plugin-dts": "2.3.0"
```
chrishelgert, line 48: vite, vitest & typescript should be then in the root package.json and also removed on the live-preview-sdk, potentially same for eslint if not already done

packages/live-preview-sdk/src/inspectorMode/utils.ts
chrishelgert: Then let's merge it as chore to not release a version for it?

packages/live-preview-sdk/src/helpers/utils.ts
```diff
@@ -189,7 +190,9 @@ export function gatherFieldInformation(
  * Parses the GraphQL query information and extracts the information,
  * we're using for processing the live updates (requested fields, alias)
  */
-export function parseGraphQLParams(query: DocumentNode): Params {
+export function parseGraphQLParams(rawQuery: DocumentNode | string): Params {
+  // If the query is a string, we need to parse it first
+  const query = typeof rawQuery === 'string' ? gql(rawQuery) : rawQuery;
```
chrishelgert, line 195: it should be already parsed in `validateConfig`

packages/live-preview-sdk/src/inspectorMode/utils.ts
chrishelgert: the next PR should be then released as canary again

2wce: yes both PRs are going to be canary releases

chrishelgert (approved):

PR #585, opened 2024-02-20
fix: re-enabling csm changes

2wce:
re-enabling the changes we disbaled in previous pr


YvesRijckaert:
@2wce did we add an eslint rule that will complain if we forget the `.js` at the end of the imports?

chrishelgert (approved):

2wce:
> @2wce did we add an eslint rule that will complain if we forget the `.js` at the end of the imports?

@YvesRijckaert no but tsc will complain if you leave it out now since we update the module resolution

PR #591, opened 2024-02-22
feat(inspectorMode): support tagging for cross space references []

chrishelgert:
Add the ability to add space and environment for inspectorMode attributes.
For the CSM they are automatically picked up

Closes #526


YvesRijckaert (approved):
Nice, thanks for taking care of this. I wonder if we should include this in the docs? Maybe only after we have the opening x-space refs in new tab functionality in the webapp

chrishelgert:
>  I wonder if we should include this in the docs? Maybe only after we have the opening x-space refs in new tab functionality in the webapp

As you mentioned, I would do it once we have also the counter part on the webapp :)

PR #592, opened 2024-02-22
fix(inspectorMode): img tagging with CSM []

chrishelgert:
Optimize tagging for images with `<picture>` or `<figure>`


YvesRijckaert (approved):
🎁

PR #593, opened 2024-02-23
fix(content-source-maps): improve tracking for richtext []

chrishelgert:
- Handle richtext correctly if all content is deeper nested
- Improve performance of content-source-map detection

**Before:**
Screenshot 2024-02-26 at 12 04 24

**Now:**
Screenshot 2024-02-26 at 12 04 56

Will create a ticket to compare the benchmarks on CI, would be nice to have this feedback (or at least failing)

### DIFF
--- a/packages/live-preview-sdk/src/inspectorMode/__tests__/getAllTaggedElements.test.ts
+++ b/packages/live-preview-sdk/src/inspectorMode/__tests__/getAllTaggedElements.test.ts
@@ -1,42 +1,15 @@
-import { SourceMapMetadata, combine } from '@contentful/content-source-maps';
+import { combine } from '@contentful/content-source-maps';
 import { describe, expect, it } from 'vitest';
 
 import { InspectorModeDataAttributes } from '../types.js';
 import { getAllTaggedElements } from '../utils.js';
+import { createSourceMapFixture } from './fixtures/contentSourceMap.js';
 
 describe('getAllTaggedElements', () => {
   const dataEntry = InspectorModeDataAttributes.ENTRY_ID;
   const dataAsset = InspectorModeDataAttributes.ASSET_ID;
   const dataField = InspectorModeDataAttributes.FIELD_ID;
   const dataLocale = InspectorModeDataAttributes.LOCALE;
-
-  function createSourceMapFixture(
-    entityId: string,
-    overrides?: {
-      origin?: string;
-      contentful?: Partial<Omit<SourceMapMetadata['contentful'], 'entity'>>;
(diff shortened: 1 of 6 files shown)


packages/live-preview-sdk/src/inspectorMode/__tests__/getAllTaggedElements.test.ts
```diff
@@ -159,6 +159,22 @@ describe('getAllTaggedElements', () => {
         expect(elements).toEqual([dom.getElementById('richtext')]);
       });
 
+      it.skip('should group sibling elements with the same information (nested structure)', () => {
```
chrishelgert, line 162: This is currently still failing, but a valid case which should work

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -127,115 +89,131 @@ function isSameElement(a: AutoTaggedElement, b: AutoTaggedElement): boolean {
...
+  let baseArray: HTMLElement[] = [];
+  if (typeof container.matches === 'function' && container.matches('*')) {
+    baseArray = [container];
+  }
 
-  // Iterate over the nodes accepted by the TreeWalker.
-  while (walker.nextNode()) {
-    const node = walker.currentNode;
+  return [...baseArray, ...Array.from(container.querySelectorAll<HTMLElement>('*'))]
+    .filter((node) => !node.matches('script, style, meta, title'))
```
z0al, line 143: You can use `container.querySelectorAll(":not(script,style,meta,title)")` instead to avoid the `.filter` call

z0al (commented):
I would be curious to see benchmarks in a real browser since vitest is using a mock DOM

packages/live-preview-sdk/src/inspectorMode/utils.ts
chrishelgert: You're right, saves some additional ms

Screenshot 2024-02-26 at 14 08 03

chrishelgert:
> I would be curious to see benchmarks in a real browser since vitest is using a mock DOM

**New version:**
- 19.599999997764826 ms
- 19.800000000745058 ms

**Before:**
- 28.299999997019768 ms
- 31.100000001490116 ms

**Difference:**
˜33% performance uplift

----

vitest benchmark (JSDOM)


**New version:**
- mean: 138 ms

**Before:**
- mean: 256 ms

**Difference:**
~45% performance uplift

YvesRijckaert (approved):

PR #609, opened 2024-03-06
feat: support editor interface and field types for CSM [TOL-1836]

YvesRijckaert:
## Description
The editor interface / field appearance of a field should determine if we generate hidden strings from the CSM for that field.

We return the editor interface as part of the CSM API GraphQL response and now have to filter it out in the SDK. This could enable future configuration for the user on the frontend side (e.g. they do want to add hidden strings to dropdown).

## Rules
- For symbol fields only create hidden strings in the SDK for the single line editor interface
- For short text list fields, we create hidden strings for all of them:
- For Rich text we always generate hidden strings
- For long text we generate hidden strings for single line, multiple line and markdown

## Todo
- [x] Before merging, we need to wait until the API side of things is done

### DIFF
--- a/packages/content-source-maps/src/__tests__/csm.spec.ts
+++ b/packages/content-source-maps/src/__tests__/csm.spec.ts
@@ -6,6 +6,26 @@ import type { SourceMapMetadata } from '../encode.js';
 import { encodeGraphQLResponse } from '../encodeSourceMap.js';
 import { GraphQLResponse } from '../types.js';
 
+const UNSUPPORTED_WIDGETS = [
+  'boolean',
+  'objectEditor',
+  'datePicker',
+  'locationEditor',
+  'rating',
+  'radio',
+  'numberEditor',
+  'urlEditor',
+  'slugEditor',
+  'dropdown',
+  'entryLinkEditor',
+  'entryCardEditor',
+  'entryLinksEditor',
+  'entryCardsEditor',
+  'assetLinkEditor',
+  'assetLinksEditor',
+  'assetGalleryEditor',
+];
+
 type Mappings = Record<string, SourceMapMetadata | Record<string, SourceMapMetadata> | undefined>;
 
 type EncodedResponse =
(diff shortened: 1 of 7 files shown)


packages/content-source-maps/src/utils.ts
```diff
@@ -15,3 +17,25 @@ export function clone<T extends Record<string, unknown> | Array<unknown>>(incomi
     return incoming;
   }
 }
+
+// We only support these editor interfaces: singleLine, tagEditor, listInput, checkbox, richTextEditor, multipleLine, markdown
+// We have to filter out the unsupported editor interfaces, instead of filtering the supported ones, because customers can add custom editor interfaces that can have any name.
+export const UNSUPPORTED_EDITOR_INTERFACES: EditorInterface[] = [
```
aodhagan-cf, line 23: Maybe it's better to do this the other way around

`SUPPORTED_EDITOR_INTERFACES` 

just to account for new editors types being added in future

aodhagan-cf (approved):
LGTM, just would recommend to flip the disallow list to an allow list

packages/content-source-maps/src/utils.ts
YvesRijckaert: Yes I had this first but commented above it why we unfortunately can't do this
Editor interfaces can be custom apps from users, which means they can have any string value (which we do want to support)

aodhagan-cf: Sorry I'm confused can you not just do the opposite filtering? also are there not custom editors we don't want to support also?

YvesRijckaert: Now we have the info if the widget is builtin or a custom one, so I changed the logic to flip the filtering 👍

packages/content-source-maps/src/__tests__/csm.spec.ts
```diff
@@ -272,13 +337,13 @@ describe('Content Source Maps', () => {
...
               {
-                title: 'Title of the post',
+                title: 'Title of the first post',
               },
               {
-                title: 'Title of the post 2',
+                title: 'Title of the second post',
               },
               {
-                title: 'Title of the post 3',
```
YvesRijckaert, line 281: Interestingly the hidden strings don't get added when there's a number at the end 🤔

chrishelgert: https://contentful.slack.com/archives/C064S7VSW5D/p1708710243893949

PR #635, opened 2024-03-20
docs: adjust gatsby notes

YvesRijckaert:
- removing gatsby example
- adjusting README


chrishelgert:
Instead of removing the Gatsby example, what do you think about just removing the live updates there?

PR #671, opened 2024-04-12
feat: cpa csm response [TOL-2011]

YvesRijckaert:
- Add support for CSM for CPA
- Remove deprecated CSM package from LP SDK, since we have it extracted in its own package now.


packages/content-source-maps/src/types.ts
```diff
@@ -49,13 +61,14 @@ export type WidgetNamespace =
...
-interface EditorInterfaceSource {
+export interface EditorInterfaceSource {
   widgetId: WidgetId;
   widgetNamespace: WidgetNamespace;
 }
 
 interface ContentSourceMaps {
-  version: number;
+  sys: { type: 'ContentSourceMaps' };
+  version: number; //@TODO remove once we remove version from GraphQL API: https://contentful.atlassian.net/browse/TOL-2016
```
chrishelgert, line 71: we could remove it directly, if more data is coming from the API it should not break 🤔

packages/content-source-maps/package.json
```diff
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@vercel/stega": "^0.1.0",
+    "contentful": "^10.8.7",
```
chrishelgert, line 43: it's only for types? then we could move it to devDependencies I think 🤔

packages/content-source-maps/src/rest/encodeCPAResponse.ts
```diff
@@ -0,0 +1,167 @@
...
+          const href = getHref(
+            entityId,
+            entityType,
+            space,
+            environment,
+            field,
+            locale,
+            targetOrigin,
+          );
+          const hiddenStrings: SourceMapMetadata = {
```
chrishelgert, line 64: we could extract this too and instead of calling getHref we call createSourceMapMetadata, then we can share this between REST and GQL too 🤔

packages/content-source-maps/src/graphql/encodeGraphQLResponse.ts
```diff
@@ -0,0 +1,91 @@
...
+  clone,
+  createSourceMapMetadata,
+  encodeField,
+  isBuiltinNamespace,
+  isSupportedWidget,
+} from '../utils.js';
+
+export const encodeGraphQLResponse = (
+  originalGraphqlResponse: GraphQLResponse,
+  targetOrigin?: 'https://app.contentful.com' | 'https://app.eu.contentful.com',
```
2wce, line 13: @YvesRijckaert do we still need to add localhost here for testing locally or is that fine?

2wce (approved):

packages/content-source-maps/src/graphql/encodeGraphQLResponse.ts
YvesRijckaert: Good question, `targetOrigin` here is only relevant for Vercel visual editor, it will open then the correct contentful link
So I don't think it's relevant to have localhost here 🤔

chrishelgert (approved):

PR #690, opened 2024-04-25
fix: refreshing page instead of redirect []

aodhagan-cf:
Refreshing window instead of calling redirect

Redirect doesn't appear to work here, and I don't believe refreshing the page is that bad.

Maybe this is relevant: https://github.com/vercel/next.js/issues/59489

### DIFF
--- a/examples/next-14-app-router-ssr/public/_live-preview.ts
+++ b/examples/next-14-app-router-ssr/public/_live-preview.ts
@@ -9,7 +9,7 @@ ContentfulLivePreview.init({
 ContentfulLivePreview.subscribe('save', {
   callback: async () => {
     const pathname = window.location.pathname;
-
-    return fetch(`/api/revalidate?pathname=${pathname}`);
+    await fetch(`/api/revalidate?pathname=${pathname}`);
+    window.location.reload();
   },
 });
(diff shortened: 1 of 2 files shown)


chrishelgert (approved):

2wce (approved):

examples/next-14-app-router-ssr/public/_live-preview.ts
```diff
@@ -9,7 +9,7 @@ ContentfulLivePreview.init({
 ContentfulLivePreview.subscribe('save', {
   callback: async () => {
     const pathname = window.location.pathname;
-
-    return fetch(`/api/revalidate?pathname=${pathname}`);
+    await fetch(`/api/revalidate?pathname=${pathname}`);
+    window.location.reload();
```
YvesRijckaert, line 13: Hey! I am struggling to understand the difference between this and the next 13 example that we reload as well?

https://github.com/contentful/live-preview/blob/main/examples/next-13-app-router-ssr/public/_live-preview.ts

Also do you mind updating the README in the examples repo to explain what the difference is now between this and the next 13 example?

aodhagan-cf: Hello! the main difference now is just the route cache invalidation request which also may or may not be needed, I was showing this to @chrishelgert and he mentioned to keep it in.

I guess when this example was originally made it assumed the redirect within the route was working. Maybe it makes sense to consolidate these examples given how close they are now?

aodhagan-cf: just this https://github.com/contentful/live-preview/blob/main/examples/next-14-app-router-ssr/app/api/revalidate/route.ts

YvesRijckaert: Yes I think having one example for SSR would be better, if they do the same thing either way :)

PR #705, opened 2024-05-02
feat: merging ssr examples in live preview []

aodhagan-cf:
Merging Next SSR examples as they have become quite similar


YvesRijckaert (approved):
thanks!

PR #713, opened 2024-05-07
fix: detect targetOrigin if not provided

MayaGillilan:


packages/live-preview-sdk/src/index.ts
```diff
@@ -133,7 +133,21 @@ export class ContentfulLivePreview {
       this.space = space;
       this.environment = environment;
 
-      this.targetOrigin = Array.isArray(targetOrigin) ? targetOrigin : [targetOrigin];
+      if (targetOrigin) {
+        this.targetOrigin = Array.isArray(targetOrigin) ? targetOrigin : [targetOrigin];
+      } else {
+        const ancestorOrigins = window.location.ancestorOrigins;
+        const currentDefaultOrigin = ancestorOrigins
+          ? DEFAULT_ORIGINS.find((origin) => ancestorOrigins.contains(origin))
```
chrishelgert, line 141: this are no longedr the default origins and instead the allowed ones?

chrishelgert: would it make sense to extract this into a function?

MayaGillilan: Hmm true, but they're still sorta defaults in that you can still override them by passing in a targetOrigin to the config. What about `ALLOWED_DEFAULT_ORIGINS`?

MayaGillilan: Yeah it's getting a bit long for a constructor so maybe a good idea

chrishelgert (approved):

PR #723, opened 2024-05-13
feat: compile csm to lower target

2wce:
this has been done to unblock [this](https://contentful.atlassian.net/browse/TOL-2080) & allow us to auto encode CPA data when  `includeContentSourceMaps` is enabled


packages/content-source-maps/vite.config.ts
```diff
@@ -9,6 +9,7 @@ export default defineConfig({
     exclude: [...configDefaults.exclude],
   },
   build: {
+    target: 'es2017',
```
chrishelgert, line 12: we have here now `es2017` as target and on `.swrc` we have `es2020`, can this lead to issues?

2wce: i think we are not using swc at all, i had made a change to the .swrc file but the build target didn't change. this must have been left over at some point

chrishelgert: maybe for a follow up, wonder if we can keep es2020 for `es` format and only use `es2017` for cjs? 🤔

chrishelgert (approved):

PR #724, opened 2024-05-13
feat: remove encode from cpa export

2wce:
as part of [this](https://contentful.atlassian.net/browse/TOL-2080?atlOrigin=eyJpIjoiZjIyOWNkYmEwZGE5NDIyYTlhZmE4YTBjZTc2MDQwOWUiLCJwIjoiaiJ9) we are now removing the encodeCPAResponse export from the LP SDK

### DIFF
--- a/packages/live-preview-sdk/src/index.ts
+++ b/packages/live-preview-sdk/src/index.ts
@@ -1,8 +1,4 @@
-import {
-  encodeGraphQLResponse,
-  encodeCPAResponse,
-  splitEncoding,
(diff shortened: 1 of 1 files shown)


z0al (changes requested):
Actual change is fine but:

- This is now a breaking a change so let's mark it accordingly in the commit message to result in a major release
- Can you please check with Yves to make sure we update customer instructions to reflect this?

YvesRijckaert (approved):
I updated the documentation so this can be merged 👍

PR #725, opened 2024-05-13
feat: sending auto tagged vs manually tagged post messages []

aodhagan-cf:
- Sending auto tagged vs manually tagged element counts
- Removing data attributes for automatically tagged elements


packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -236,10 +240,19 @@ export function getAllTaggedElements(root = window.document, ignoreManual?: bool
     element.setAttribute(InspectorModeDataAttributes.LOCALE, sourceMap.contentful.locale);
     element.setAttribute(InspectorModeDataAttributes.SPACE, sourceMap.contentful.space);
     element.setAttribute(InspectorModeDataAttributes.ENVIRONMENT, sourceMap.contentful.environment);
+    element.setAttribute(InspectorModeDataAttributes.AUTO_TAGGED, 'true');
     taggedElements.push(element);
   }
 
-  return taggedElements;
+  const autoTaggedCount = Array.from(manualTagged).filter((el) =>
+    el.hasAttribute(InspectorModeDataAttributes.AUTO_TAGGED),
```
chrishelgert, line 248: instead of auto-tagged could we add a property for manual tagging (similar to vercel is doing it for their visual editing)

Caveat: We would need to check for the SDK version (before the current version it was always manual tagged)

aodhagan-cf: Yup I'll adjust this now

packages/live-preview-sdk/src/inspectorMode/types.ts
```diff
@@ -61,6 +64,8 @@ export type InspectorModeMouseMoveMessage = {
 };
 export type InspectorModeTaggedElementsMessage = {
   elements: Array<InspectorModeElement>;
+  manualTagCount: number;
```
chrishelgert, line 67: NIT: should it be here manuallyTaggedCount? (and automaticallyTaggedCount)
we have a mix of manualTag and manuallyTagged 🤷

aodhagan-cf: I've adjust to `manuallyTaggedCount` and `automaticallyTaggedCount` for consistency

aodhagan-cf: I've adjusted

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -239,7 +243,15 @@ export function getAllTaggedElements(root = window.document, ignoreManual?: bool
...
   }
 
-  return taggedElements;
+  const autoTaggedCount = Array.from(manualTagged).filter(
+    (el) => !el.hasAttribute(InspectorModeDataAttributes.MANUALLY_TAGGED),
+  ).length;
+
+  return {
+    taggedElements,
+    manuallyTaggedCount: manualTagged.length - autoTaggedCount,
```
z0al, line 252: Umm.. why not manualTagged.length?

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -51,6 +51,7 @@ export function getInspectorModeAttributes(
     environment:
       element.getAttribute(InspectorModeDataAttributes.ENVIRONMENT) ?? fallbackProps.environment,
     space: element.getAttribute(InspectorModeDataAttributes.SPACE) ?? fallbackProps.space,
+    manuallyTagged: element.hasAttribute(InspectorModeDataAttributes.MANUALLY_TAGGED),
```
z0al, line 54: Since we convert the encoded strings to normal data attributes but without the new manually tagged flag and in [line 246 below](https://github.com/contentful/live-preview/pull/725/files#diff-7371daf98a723e8f1567e59dd41ac1acfe948cfc488c6cbf4ede34b6e4213021R246-R248) you are only checking that the new attribute doesn't exist. 

Wouldn't this report auto-tagged as manually tagged if called twice on the dom? I know we probably don't care about its value anymore but still feels wrong.

Maybe doing it the other way around would be better, like adding `data-contentful-auto-tagged=true` or `data-contentful-tagging="auto"`.

aodhagan-cf: I did it the other way around first and @chrishelgert suggested the opposite. I originally had the data attribute as auto and changed it to manual. This was to align with to how vercel does it: https://github.com/contentful/live-preview/pull/725#discussion_r1599702366

packages/live-preview-sdk/src/inspectorMode/utils.ts
aodhagan-cf: I think I should rename `manualTagged` is it isn't just manual tagged. It's actually simply `tagged` as it just picks up the elements with the tagging data attributes. (was like this before the PR)

The logic is confusing I know, I'll tidy it up and add comments. But to clarify this is how it is now

- manualTagged (all elements that have tagging data attributes)
- uniqElementsForTagging are the elements that we have just added the autotagging to
- autoTaggedCount = manualTagged - the manualTagged that don't have the manual data attribute

packages/live-preview-sdk/src/inspectorMode/utils.ts
aodhagan-cf: > Wouldn't this report auto-tagged as manually tagged if called twice on the dom? I know we probably don't care about its value anymore but still feels wrong.

I would say manual tagging should take priority over autotagging

packages/live-preview-sdk/src/inspectorMode/utils.ts
aodhagan-cf: I've simplified the code thanks for calling it out

packages/live-preview-sdk/src/inspectorMode/utils.ts
chrishelgert: Additional thought: 
Longterm we will have more auto-tagged than manual tagged fields (🤞 )
So adding it on manual fields should lead to a smaller fingerprint, than on auto-tagged fields

aodhagan-cf: @z0al 
> To fix the current way I think we need to
As I see it we have 2 states, manually tagged or auto tagged so if if it has the attribute then it's manually tagged if it doesn't then it's auto tagged. How would your suggestion fix what is currently here?

aodhagan-cf: Also agree with @chrishelgert here that just tagging the manual has a smaller footprint

chrishelgert (approved):
LGTM

PR #727, opened 2024-05-17
fix(inspectorMode): more accurate outlines [TOL-1718]

chrishelgert:
- [x] Update on mouse move
- [x] Check for visibility of elements
- [x] fix element hover
- [x] Fix typings for `useInspectorMode`
- [x] cleanup (don't send the same message twice)
- [x] (Stretch: refresh on webpack hmr)

### DIFF
--- a/packages/live-preview-sdk/src/index.ts
+++ b/packages/live-preview-sdk/src/index.ts
@@ -154,6 +154,7 @@ export class ContentfulLivePreview {
 
       // bind event listeners for interactivity
       window.addEventListener('message', (event: MessageEvent<MessageFromEditor>) => {
+        console.log('>> message', event);
         if (!isValidMessage(event)) {
           return;
         }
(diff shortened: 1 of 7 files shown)


packages/live-preview-sdk/src/index.ts
```diff
@@ -154,6 +154,7 @@ export class ContentfulLivePreview {
 
       // bind event listeners for interactivity
       window.addEventListener('message', (event: MessageEvent<MessageFromEditor>) => {
+        console.log('>> message', event);
```
YvesRijckaert, line 157: 👓

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -96,38 +135,38 @@ export class InspectorMode {
...
-      if (this.taggedElements?.length !== taggedElements.length) {
-        this.sendAllElements();
-      }
+      this.updateElements();
     });
-
-    mutationObserver.observe(document.body, {
+    mutationObserver.observe(element, {
+      // Content source maps
+      characterData: true,
```
YvesRijckaert, line 155: What does this have to do exactly with CSM?

chrishelgert: content source maps is only characters, so If the characters changes there could be new tagged elements
Thats why we have to check for characterData :)

YvesRijckaert (approved):

PR #732, opened 2024-05-23
fix: inspector mode types [ZEND-5004]

YvesRijckaert:
When using `"strictNullChecks": false` in `tsconfig.json` using a project with the live preview SDK and inspector mode, you would get this error:

```
Object literal may only specify known properties, and 'fieldId' does not exist in type '({ entryId: string; } | { assetId: string; }) & Pick<LivePreviewProps, "locale" | "space" | "environment">'.
```

I simplified the `GetInspectorModeProps` type which fixes this issue

### DIFF
--- a/packages/live-preview-sdk/src/react.tsx
+++ b/packages/live-preview-sdk/src/react.tsx
@@ -17,13 +17,7 @@ import isEqual from 'lodash.isequal';
 
 import { debounce } from './helpers/index.js';
 import { ContentfulLivePreview, ContentfulLivePreviewInitConfig } from './index.js';
-import type { InspectorModeTags } from './inspectorMode/types.js';
-import {
-  Argument,
-  LivePreviewAssetProps,
-  LivePreviewEntryProps,
-  LivePreviewProps,
-} from './types.js';
+import { Argument, LivePreviewProps } from './types.js';
 
 type UseEffectParams = Parameters<typeof useEffect>;
 type EffectCallback = UseEffectParams[0];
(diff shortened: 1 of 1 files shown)


packages/live-preview-sdk/src/__tests__/react.spec.tsx
```diff
@@ -409,7 +414,7 @@ describe('useContentfulInspectorMode', () => {
     });
 
     it('should return null because the inspector mode is disabled', () => {
-      const { result } = renderHook((data) => useContentfulInspectorMode(data), {
+      const { result } = renderHook((data) => useContentfulInspectorMode<any>(data), {
```
chrishelgert, line 417: this any seems like another issue? 🤔

YvesRijckaert: I think it's fine? Because otherwise it complains that we don't pass certain fields in
```
result.current({
  locale,
  fieldId: 'title',
}),
```

e.g. we don't pass `entryId` but pass it `initialProps`

packages/live-preview-sdk/src/react.tsx
```diff
@@ -191,20 +191,7 @@ export function useContentfulLiveUpdates<T extends Argument | null | undefined>(
...
-        >]: LivePreviewEntryProps[K];
-      }
-    | {
-        [K in Exclude<
-          keyof LivePreviewAssetProps,
-          keyof T | 'locale' | 'environment' | 'space'
-        >]: LivePreviewAssetProps[K];
-      }
-  ) &
+  props: ((LivePreviewEntryProps | LivePreviewAssetProps) & T) &
```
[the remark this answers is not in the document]
YvesRijckaert: If I implement this change, then it wouldn't complain about a missing `fieldId`

packages/live-preview-sdk/src/__tests__/react.spec.tsx
chrishelgert: Screenshot 2024-05-23 at 15 22 16

if we pass entryId on the initialProps it fails on the second call as it expect it to be provided there too

chrishelgert (approved):

PR #753, opened 2024-06-03
fix: autotagging attributes []

YvesRijckaert:


packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -191,7 +208,13 @@ export function getAllTaggedElements(
         `[${InspectorModeDataAttributes.ASSET_ID}][${InspectorModeDataAttributes.FIELD_ID}], [${InspectorModeDataAttributes.ENTRY_ID}][${InspectorModeDataAttributes.FIELD_ID}]`,
       );
 
-  const taggedElements: Element[] = [...alreadyTagged];
+  //Spread operator is necessary to convert the NodeList to an array
+  const taggedElements: TaggedElement[] = [...alreadyTagged]
+    .map((element: Element) => ({
+      element,
+      attributes: getInspectorModeAttributes(element, options),
```
chrishelgert, line 215: we could add a property `manuallyTagged: true` here, then we don't need the whole checks on Line 282 :)

packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -19,7 +19,7 @@ type InspectorModeOptions = {
 };
 
 type TaggedElement = {
-  attributes: InspectorModeAttributes;
+  attributes: InspectorModeAttributes | null;
```
chrishelgert, line 22: why do we need `null` here?
Elements without attributes should not be tagged? 🤔

packages/live-preview-sdk/src/inspectorMode/utils.ts
YvesRijckaert: It already gets added by getInspectorModeAttributes :D I've also renamed this function to make it more clear that this is for manually tagged inspector mode

packages/live-preview-sdk/src/inspectorMode/index.ts
YvesRijckaert: `getManualInspectorModeAttributes` sets null sometimes:

>  * Parses the necessary manually tagged information from the element and returns them.
>  * If **one** of the information is missing it returns null

chrishelgert (approved):

PR #754, opened 2024-06-07
feat: add-adjustments-to-pointers [TOL-2036]

2wce:
adjusting the changes made in the graphql api & CPA. This will be a breaking change & will be reflected as such in the squash commit


z0al:
@2wce I think this is a breaking change but on the API side on the SDK

z0al (approved):

PR #785, opened 2024-07-05
feat(content-source-maps): allow specify a platform to reduce the payload []

chrishelgert:
Provides the ability to reduce the payload of the hidden strings if only one platform for previewing is used
- using `contentful` - removing the `href` property
- using `vercel` - removing the `contentful` property

### DIFF
--- a/packages/content-source-maps/src/__tests__/graphql.spec.ts
+++ b/packages/content-source-maps/src/__tests__/graphql.spec.ts
@@ -880,6 +880,156 @@ describe('Content Source Maps with the GraphQL API', () => {
     });
   });
 
+  test('adds only the `href` parameter in the encoding if platform `vercel` is configured', () => {
+    const graphQLResponse: GraphQLResponse = {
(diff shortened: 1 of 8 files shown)


z0al:
> Provides the ability to reduce the payload of the hidden strings if only one platform for previewing is used
> 
> * using `contentful` - removing the `href` property
> * using `vercel` - removing the `contentful` property

How would this now behave if Vercel is chosen but the website is still opened within Contentful Live Preview?

chrishelgert:
> > Provides the ability to reduce the payload of the hidden strings if only one platform for previewing is used
> > 
> > * using `contentful` - removing the `href` property
> > * using `vercel` - removing the `contentful` property
> 
> How would this now behave if Vercel is chosen but the website is still opened within Contentful Live Preview?

then you would have no tagged elements in live preview. this is really if you want to opt-out one the preview functionalities

z0al (approved):

Issue #820, opened 2024-07-30
Can't Import the named export 'version' (imported as 'version') from default-exporting module


tobslob:
**Describe the bug**
The latest release of v4.4.0 throws this error, I had to downgrade to v4.3.1

To replicate, update the preview url using localhost, and update to the latest @contentful/live-preview 4.4.0

<img width="1057" alt="image" src="https://github.com/user-attachments/assets/43b1bd88-3523-4678-b3bc-b1a1d830da0b">

2wce:
this should now be fixed by #821 @tobslob , thank you for reporting this 👍🏼

PR #821, opened 2024-07-30
fix: import package version correctly

2wce:
previous version is broken possibly due to lower module target. this fixes it


2wce:
fixes #820

chrishelgert (approved):

tobslob:
This issue still persist.

PR #823, opened 2024-07-31
fix(import/export): import version correctly on the remaining place []

chrishelgert:
follow up on https://github.com/contentful/live-preview/pull/821


2wce (approved):
thanks for picking this up

PR #846, opened 2024-08-21
docs(known issues): add genereal block about embedding into another iframe than contentful app []

chrishelgert:
https://discord.com/channels/1030176999471321129/1268975295667372073


YvesRijckaert (approved):
Nice one, can we also add it to the doc-app repo?

PR #893, opened 2024-09-18
fix: improve error messaging

MayaGillilan:
Makes debug messaging more explicit and meaningful, so as to make configuration issues and user debugging easier

### DIFF
--- a/packages/live-preview-sdk/src/index.ts
+++ b/packages/live-preview-sdk/src/index.ts
@@ -345,6 +345,7 @@ export class ContentfulLivePreview {
     return this.liveUpdatesEnabled;
   }
 
+  //TODO: Rename to openEntityInEditor, as assets can also be opened. Would be breaking change
   static openEntryInEditor(props: LivePreviewProps): void {
     const defaultProps = {
       locale: this.locale,
@@ -374,7 +375,7 @@ export class ContentfulLivePreview {
       return;
     }
 
-    debug.error('Please provide field id and entry/asset id to openEntryInEditor.');
+    debug.error('Please provide field id and entry/asset id to openEntryInEditor.', { ...props });
   }
 
   /**
--- a/packages/live-preview-sdk/src/index.ts
+++ b/packages/live-preview-sdk/src/index.ts
@@ -92,7 +92,7 @@ export class ContentfulLivePreview {
   static init(config: ContentfulLivePreviewInitConfig): Promise<InspectorMode | null> | undefined {
     if (typeof config !== 'object' || !config?.locale) {
       throw new Error(
-        "Init function have to be called with a locale configuration (for example: `ContentfulLivePreview.init({ locale: 'en-US'})`)",
+        "Init function has to be called with a locale configuration (for example: `ContentfulLivePreview.init({ locale: 'en-US'})`)",
       );
     }
 
@@ -374,7 +374,7 @@ export class ContentfulLivePreview {
       return;
     }
 
-    debug.error('Please provide field id and entry id to openEntryInEditor.');
+    debug.error('Please provide field id and entry/asset id to openEntryInEditor.');
   }
 
   /**
--- a/packages/live-preview-sdk/src/inspectorMode/utils.ts
+++ b/packages/live-preview-sdk/src/inspectorMode/utils.ts
@@ -68,7 +68,7 @@ export function getManualInspectorModeAttributes(
   };
 
   if (!sharedProps.fieldId) {
-    debug.warn('Tagged element is missing field ID attribute and cannot be tagged', {
+    debug.warn('Element is missing field ID attribute and cannot be tagged', {
       id:
         element.getAttribute(InspectorModeDataAttributes.ENTRY_ID) ??
         element.getAttribute(InspectorModeDataAttributes.ASSET_ID),
--- a/packages/live-preview-sdk/src/inspectorMode/utils.ts
+++ b/packages/live-preview-sdk/src/inspectorMode/utils.ts
@@ -67,6 +67,16 @@ export function getManualInspectorModeAttributes(
     manuallyTagged: true,
   };
 
+  if (!sharedProps.fieldId) {
+    debug.warn('Tagged element is missing field ID attribute and cannot be tagged', {
+      id:
+        element.getAttribute(InspectorModeDataAttributes.ENTRY_ID) ??
+        element.getAttribute(InspectorModeDataAttributes.ASSET_ID),
+      sharedProps,
+    });
+    return null;
+  }
+
   const entryId = element.getAttribute(InspectorModeDataAttributes.ENTRY_ID);
   if (entryId) {
     return { ...sharedProps, entryId };
@@ -148,18 +158,16 @@ function hasTaggedParent(node: HTMLElement, taggedElements: TaggedElement[]): bo
 export function getAllTaggedElements({
   root = window.document,
   options,
-  ignoreManual,
 }: {
   root?: any;
   options: Omit<InspectorModeOptions, 'targetOrigin'>;
-  ignoreManual?: boolean;
 }): {
   taggedElements: TaggedElement[];
   manuallyTaggedCount: number;
   automaticallyTaggedCount: number;
   autoTaggedElements: AutoTaggedElement<Element>[];
 } {
-  const alreadyTagged = ignoreManual
+  const alreadyTagged = options.ignoreManuallyTaggedElements
     ? []
     : root.querySelectorAll(
         `[${InspectorModeDataAttributes.ASSET_ID}][${InspectorModeDataAttributes.FIELD_ID}], [${InspectorModeDataAttributes.ENTRY_ID}][${InspectorModeDataAttributes.FIELD_ID}]`,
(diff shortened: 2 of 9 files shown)


packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -180,7 +188,13 @@ export function getAllTaggedElements({
   for (const { node, text } of stegaNodes) {
     const sourceMap = decode(text);
     if (!sourceMap || !sourceMap.origin.includes('contentful.com')) {
-      debug.warn;
+      debug.warn(
+        "Element has missing or invalid ContentSourceMap, please check if you have correctly enabled ContentSourceMaps and that the element's data originates from Contentful",
```
2wce, line 192: are we trying to imply that the data shouldn't be transformed once it comes from our APIs?

MayaGillilan: I have to admit I'm not as familiar with CSM so that was my understanding, but if that's an inaccurate description please let me know what would be more accurate

packages/live-preview-sdk/src/index.ts
```diff
@@ -374,7 +374,7 @@ export class ContentfulLivePreview {
       return;
     }
 
-    debug.error('Please provide field id and entry id to openEntryInEditor.');
+    debug.error('Please provide field id and entry/asset id to openEntryInEditor.');
```
YvesRijckaert, line 377: Can we provide an asset id to `openEntryInEditor` ? 🤔 Shouldn't it be renamed then to `openEntityInEditor`

MayaGillilan: We can indeed, and that would make more sense, although I wonder if it would be a breaking change 🤔

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -67,6 +67,16 @@ export function getManualInspectorModeAttributes(
     manuallyTagged: true,
   };
 
+  if (!sharedProps.fieldId) {
+    debug.warn('Tagged element is missing field ID attribute and cannot be tagged', {
```
aodhagan-cf, line 71: This reads a bit weird, as in how can a tagged element not be tagged 😅

MayaGillilan: That's very true 😅 I'll try to think of a different wording

packages/live-preview-sdk/src/inspectorMode/utils.ts
MayaGillilan: @2wce @YvesRijckaert Can you confirm if this is an appropriate message?

packages/live-preview-sdk/src/index.ts
MayaGillilan: Added a todo since it would be a breaking change, and also added the props to this error

packages/live-preview-sdk/src/inspectorMode/utils.ts
aodhagan-cf: I guess you could also say something like "Element is missing field ID attribute so it cannot be tagged"

packages/live-preview-sdk/src/liveUpdates.ts
```diff
@@ -34,6 +34,17 @@ export class LiveUpdates {
 
       const subscription = this.subscriptions.get(subscriptionId);
 
+      if (!data.sys || !data.sys.id) {
```
YvesRijckaert, line 37: @MayaGillilan Just as an FYI: this broke live updates for GraphQL:
`data` could look like this for GraphQL:

```
{
    "postCollection": {
        "items": [
            {
                "__typename": "Post",
                "sys": {
                    "id": "a1b2c3"
                },
                "slug": "foo",
                "title": "new value​​​​‌﻿‍﻿​‍​‍‌‍﻿﻿‌﻿​‍‌‍‍‌‌‍‌﻿‌‍‍‌‌‍﻿‍​‍​‍​﻿‍‍​‍​‍‌‍​﻿‌‍﻿﻿‌‍﻿‍‌﻿‌​‌‍‌‌‌‍﻿‍‌﻿‌​‌‍‌‍‌﻿‌‌‌‍﻿​​‍﻿‍‌‍​﻿‌‍﻿﻿‌‍﻿‌​‍​‍​‍﻿​​‍​‍‌‍‍​‌﻿​‍‌‍‌‌‌‍‌‍​‍​‍​﻿‍‍​‍​‍‌‍‍​‌﻿‌​‌﻿‌​‌﻿​​‌﻿​﻿​﻿‍‍​‍﻿﻿​‍﻿﻿‌‍​‌‌﻿​​‌﻿​​​‍﻿‍‌‍​﻿‌‍﻿﻿‌‍﻿‍‌﻿‌​‌‍‌‌‌‍﻿‍‌﻿‌​‌‍‌‍‌﻿‌‌‌‍﻿​​‍﻿‍‌‍​﻿‌‍﻿﻿‌‍﻿‌​‍﻿﻿‌﻿​﻿‌﻿​​‌‍​‌‌‍​﻿‌‍‌‌‌﻿​﻿​‍﻿﻿‌﻿‌​‌‍‌‌‌﻿​﻿‌﻿‌​​‍﻿‌‌﻿​﻿‌﻿​​‌‍​‌‌‍​﻿‌‍‌‌​‍﻿﻿‌‍‌‌‌‍﻿‍‌﻿‌‍‌‍‍‌‌﻿​‍‌‍﻿﻿‌‍﻿‍‌‍﻿‌‌‍‌‌‌‍﻿‍‌﻿‌​‌﻿​﻿​‍﻿﻿‌‍﻿‌‌‍​‌‌﻿​﻿‌﻿‌​‌‍‌‌‌﻿​‍​‍﻿﻿‌‍‌‌‌‍﻿‍‌﻿‌​‌﻿​‍‌‍‍‌‌‍‌‌‌﻿​﻿​‍﻿﻿‌‍​‌​﻿​‌‌‍​‍​﻿​‍‌‍​﻿​﻿​﻿​‍﻿﻿​﻿﻿﻿‌‍‌‍‌‍﻿﻿‌‍​﻿‌﻿‌‌‌﻿​﻿‌‍‌‌‌‍‌​‌​‌‍‌‍‍‌‌‍‌‌‌‍﻿​‌‍‌​​﻿﻿‌‌﻿‌​‌‍‍‌‌﻿‌​‌‍﻿​‌‍‌‌​‍‌‍‌‍‌‍‌‍﻿﻿‌‍​﻿‌﻿‌‌‌﻿​﻿‌‍‌‌‌‍‌​‌​﻿​‌‍﻿﻿‌‍​﻿‌‍​‌‌‍﻿​‌‍‌‌​﻿﻿‌‌‍‌‌‌‍﻿‍​‍﻿‌‌‌‌‌‌‌​﻿​‍‌‍‌﻿​﻿‌‍﻿﻿‌﻿‌‌‌﻿​‍‌‍​﻿‌‍‌‌​﻿﻿‌‌﻿‌‍‌‍‌‌‌﻿​‍‌‍​﻿‌‍‌‌‌‍﻿​​‍﻿‌‌‍​﻿‌‍﻿﻿‌‍﻿‍‌﻿‌​‌‍‌‌‌‍﻿‍‌﻿‌​​‍﻿‌‌‍﻿​‌‍‍‌‌‍﻿‍‌‍‍﻿​‍​‍​‍﻿​​‍​‍‌‍​﻿‌‍﻿﻿‌‍﻿‍‌﻿‌​‌‍‌‌‌‍﻿‍‌﻿‌​‌‍‌‍‌﻿‌‌‌‍﻿​​‍​‍​﻿‍‍‌﻿‍﻿​‍​‍‌‍‌‌‌‍‌​‌‍‍‌‌﻿‌​‌‍﻿﻿‌﻿​‍‌​‍‌‌‍﻿‍‌﻿‌​‌‍‌‌‌﻿​‍‌‍‌‍‌‍​‌‌‍​﻿‌‍‌‌​‍​‍​﻿‍‍‌﻿‍﻿​‍​‍‌﻿‌﻿‌‍‍‌‌‍‌​‌‍‌﻿‌‍‌‌‌﻿‌​‌​﻿‍‌‍​‌‌‍﻿‌‌‍‌‌‌﻿​﻿‌﻿​​‌‍​‌‌‍​﻿‌‍‌‌​‍​‍​﻿‍‍​‍​‍‌‍​‍‌﻿‌‌‌‍‍‌‌‍﻿​‌﻿‌​‌‍‍‌‌‍﻿‍​‍​‍​‍﻿​​‍​‍‌﻿‌﻿‌‍‍‌‌‍‌​‌‍‌﻿‌‍‌‌‌﻿‌​‌​‍‌‌‍‌​​‍​‍​﻿‍‍​‍​‍‌﻿​﻿‌‍‍‌‌‍﻿‍‌‍‌﻿‌‍﻿​‌‍‌‌‌​﻿​‌‍‍‌‌‍﻿‍‌‍‌‌​‍​‍‌﻿﻿‌​‍﻿​​‍​‍‌‍‌‍‌‍‍‌‌‍‌‌‌‍﻿​‌‍‌​‌‌‌​‌﻿‍‌‌﻿​​‌‍‌‌​‍​‍​﻿‍‍​‍​‍‌‌​﻿‌﻿‍‌‌‍﻿‌‌‍​‍‌‍﻿﻿‌‍﻿​​‍​‍‌﻿﻿‌‌﻿﻿‌",
                ]
            }
        ]
    }
```

YvesRijckaert: https://github.com/contentful/live-preview/pull/926

PR #894, opened 2024-09-18
fix: adding new property to tagged elements that are overlapped by others [TOL-2408]

aodhagan-cf:
Checking intersections of overlapping tagged elements and adding an attribute which is sent to the editor to stop rendering tagged elements which should be hidden.

- For each 4 corners of the tagged elements calculate what the topmost element is at that point
- If the topmost element is not the tagged element add property to hide tagged element

### DIFF
--- a/packages/live-preview-sdk/src/inspectorMode/index.ts
+++ b/packages/live-preview-sdk/src/inspectorMode/index.ts
@@ -319,28 +319,42 @@ export class InspectorMode {
     });
   };
 
+  private getAllLayersInDocument = (): { zIndex: number; coordinates: DOMRect }[] => {
+    const allElements = document.querySelectorAll('*');
+    const layers = [];
+    for (const element of allElements) {
+      const zIndex = this.getZIndex(element);
+      if (zIndex !== 'auto' && zIndex !== null) {
+        layers.push({
+          zIndex: Number(zIndex),
+          coordinates: element.getBoundingClientRect(),
+        });
+      }
+    }
+    return layers;
+  };
+
   /**
    * 1. Checks if the element is visible
-   * 2. Checks if the element is covered by another element with a higher z-index
+   * 2. Checks if the element is covered by a layer of higher z-index
    */
-  private addVisibilityAttributesToTaggedElements = (taggedElements: Partial<TaggedElement>[]) =>
-    taggedElements.map((taggedElement, currentIndex) => {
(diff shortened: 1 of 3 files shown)


packages/live-preview-sdk/src/inspectorMode/index.ts
```diff
@@ -239,10 +235,11 @@ export class InspectorMode {
       {
         elements: this.taggedElements.map((taggedElement) => ({
           // Important: do not add `element` as it can't be cloned by sendMessage
-          coordinates: taggedElement.coordinates,
-          isVisible: taggedElement.isVisible,
+          coordinates: taggedElement.coordinates!,
+          isVisible: !!taggedElement.isVisible,
```
chrishelgert, line 239: Could we fix the typings when we create the taggedElements?

chrishelgert (approved):

packages/live-preview-sdk/src/inspectorMode/index.ts
aodhagan-cf: I've added a new type now

PrecaulculatedTaggedElement 

This just has the stuff that you can get directly from the tagged element, so element + attributes

The the actual TaggedElement has the calculated state without any conditional properties

packages/live-preview-sdk/src/inspectorMode/utils.ts
```diff
@@ -271,3 +281,53 @@ export function getAllTaggedEntries({
...
+  }));
+
+/**
+ * Applies the attributes that we cannot simply get from the tagged elements itself
+ * but need to calculate based on the current state of the document
+ */
+export const addCalculatedAttributesToTaggedElements = (
+  taggedElements: PrecaulculatedTaggedElement[],
+  root = window.document,
+): TaggedElement[] => {
```
chrishelgert, line 330:
```suggestion
): PrecaulculatedTaggedElement[] => {
```

chrishelgert: same then in 332

aodhagan-cf: No this should use the regular one as it now has added the calculations

aodhagan-cf: I've merged for now, let me know I can follow up later on this if you disagree

PR #955, opened 2024-11-21
chore: update ssr example with refs []

2wce:
updating the ssr example to show how to tag a nested reference

### DIFF
--- a/examples/next-app-router-ssr/lib/api-graphql.ts
+++ b/examples/next-app-router-ssr/lib/api-graphql.ts
@@ -11,9 +11,7 @@ export interface Post {
   banner: {
     sys: Sys;
     title: string;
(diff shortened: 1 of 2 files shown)


examples/next-app-router-ssr/lib/api-graphql.ts
```diff
@@ -21,13 +28,22 @@ interface FetchResponse {
...
+  }
+  slug
+  title
+  description
+  banner {
+    sys {
+      id
+    }
+    title
+    file {
```
chrishelgert, line 43: just double-checked here: https://www.contentful.com/developers/docs/references/graphql/
it should be directly URL without `file`

chrishelgert (approved):

Issue #858, opened 2024-08-27
Differentiating between entry update types


neilvanbeinum:
Hello,

We would like to integrate live previewing with our client app.

We would like to perform a simple page reloading to achieve this, as opposed to updating the DOM. We are using vanilla JS.

The issue we're facing is that the handler passed to the `ContentfulLivePreview.subscribe` function is called each time the page is reloaded. We would need to introduce a delay following reload to prevent an infinite reload loop. We'd prefer to avoid the delay.

Is there a way (either current or planned) to differentiate between updates fired from an entry update "directly" and entry updates following a page reload?

Thanks,

Neil

chrishelgert:
👋 
Do you subscribe to the `save` or `edit` (default) event?
The `edit` event replays the last changes again - which could trigger the mentioned infinite loop, at the moment we have nothing to avoid this.
Maybe the `save` event is what you could use for this approach (Note: the save event is coupled to the inspectorMode at the moment)

chrishelgert:
Created a ticket for us internally too, to give the opportunity to disable the tagged element check

PR #1027, opened 2025-01-22
fix: trigger save event also when inspector mode is disabled

MayaGillilan:


packages/live-preview-sdk/src/saveEvent.ts
```diff
@@ -42,10 +52,15 @@ export class SaveEvent {
...
+      if (this.inspectorModeEnabled) {
+        const entries = getAllTaggedEntries({ options: this.options });
 
-      if (entries.includes(entity.sys.id)) {
+        if (entries.includes(entity.sys.id)) {
+          this.subscription(entity);
+        }
+      } else {
         this.subscription(entity);
+        window.location.reload();
```
MayaGillilan, line 63: Note: Do we want to hardcode this or have users provide it in the callback if they wish for reloading?

chrishelgert: user have to provide it - we should have an example in the examples directory for it

chrishelgert: the reason for providing it, you could have a reload implementation by the framework you're using for routing (e.g. next.js, react-router, ..)

MayaGillilan: Makes sense, I'll remove this line then!

chrishelgert (approved):

chrishelgert:
implements https://github.com/contentful/live-preview/issues/858

Issue #1100, opened 2025-03-25
Inspector data becomes outdated when using `useContentfulLiveUpdates` & GraphQL


adrosouthpaw:
### Using
- Remix/RR7 - `react-router@7.4.0`
- Contentful Live Preview - `@contentful/live-preview@4.6.11`

### Code

```tsx
import { encodeGraphQLResponse } from "@contentful/live-preview";
import { useContentfulLiveUpdates } from "@contentful/live-preview/react";
import { useLoaderData } from "react-router";

export async function loader() {
  // calls and returns the RAW GraphQL data from Contentful
  // Query contains `@contentSourceMaps`
  // Using `preview: true` and the preview token
}

export default function Page() {
  // Get the raw data from the Server
  const result = useLoaderData<typeof loader>();
  // Pass it to the live updates hook to get new data as needed
  const updated = useContentfulLiveUpdates(result.data);
  // Encoded the new data with the GraphQL extension data
  const encoded = encodeGraphQLResponse({
    data: updated as any,
    /*
      FIXME:
      I think the issue is here, the live updates hook doesn't provide
      updated extension data for the encoding to be accurate
    */
    extensions: result.extensions,
  });

  //... passes the `encoded` data to components to render
}
```

### Problems

So the live updates do work in general however, there are a few instances where it doesn't do what you'd expect.
1. Adding a new reference (Flexible Assembly setup) - this **will** show up but it **won't** have any inspect controls until you reload the live preview window.
2. Switching 2 records of **different** types will result in the inspect controls disappearing from both, until you reload the live preview window.
3. Switching 2 records of the **same** type will keep the inspect controls, but the "target" won't change. So if you go from `A B` to `B A` and edit `B`, it will open the editor for `A`.

https://github.com/user-attachments/assets/2abb270e-7aee-421c-bc84-56148d20b655

### Possible problem area

I think this is happening because you're only getting the extensions data on the initial page load and never again.

### Possible solution

I've tried to pass the query I use on the server to the `query` option in the `useContentfulLiveUpdates` hook. However, that hook will only ever return the data. It would be good if, once you provide a query the return type changes and it sends back both `data` and `extensions`. That way you can keep the `encodeGraphQLResponse` aligned with correct extensions data.

I don't think a acceptable solution here would be to switch to the "Manual Tagging" implementation to support Live Preview Inspect. Especially since the docs "strongly recommend Content Source Maps option since tagging is automated".

YvesRijckaert:
[...]

PR #1110, opened 2025-04-01
docs: add explanation about CSM and live updates [TOL-2936]

YvesRijckaert:
- Add section on combining live updates + content source maps
- Update example

Fixes: https://github.com/contentful/live-preview/issues/1100

### DIFF
--- a/examples/content-source-maps-cpa/src/pages/index.tsx
+++ b/examples/content-source-maps-cpa/src/pages/index.tsx
@@ -1,18 +1,20 @@
 import { GetStaticProps } from 'next';
-import { Post, getAllPostsForHome } from '../../lib/api-rest';
+import { Post as PostType, getAllPostsForHome } from '../../lib/api-rest';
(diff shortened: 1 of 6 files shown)


examples/content-source-maps-cpa/src/pages/index.tsx
```diff
@@ -1,21 +1,21 @@
 import { GetStaticProps } from 'next';
-import { Post, getAllPostsForHome } from '../../lib/api-rest';
+import { Post as PostType, getAllPostsForHome } from '../../lib/api-rest';
+import { useContentfulLiveUpdates } from '@contentful/live-preview/react';
+
+const Post = ({ post }: { post: PostType }) => {
+  const updatedPost = useContentfulLiveUpdates(post);
```
chrishelgert, line 6: you will only get updates for post, but not if a new one is added should we move it one level up?

chrishelgert (approved):
One tiny improvement

PR #1175, opened 2025-06-12
docs: rename to rsc [TOL-3135]

YvesRijckaert:
Rename from SSR to RSC to make it more clear that we're dealing with React Server Components


2wce (approved):
LGTM

